PR #11, opened 2023-10-19
Add RS title screens

Very-Soft:


Very-Soft:
Oh Kenzie already did this lol

PR #210, opened 2024-04-09
Sorted Markings Selection for Preference menu

Poojawa:
# Only legacy uncolorable ears and tails like Squirrel, Red and Wise Wolf  were removed Path wise. 

This really needs to have a quality test run like the 515 testing on people with complex markings to ensure nothing is broken. Mostly to cut down the potential explosion of angy.

🆑 Poojawa
rscadd: Provided a means to choose markings based on category.
/:cl:

928c74709bfb6c750321835e17c91f8b
d759eb68b35a273d61c1d69423e99d2f

I did not touch the paths, but if I was able to, I'd have made this a little less brute-force. but alas.

Allows players at character creation to browse a specific category of markings, rather than all of them in one full list. Filters Teshari specifics out as well so Teshari-only characters can easily see Teshari-only things like the scars or bandages.


Very-Soft (changes requested):
So, one thing I noticed basically immediately with this change in testing, is that options you have already selected are able to be selected again, which will reset any colors you have placed on them.

Further, the fingerless socks, and fingers and toes options appear inside of the head and face category

Additionally, it seems like some of the lists have a 'Normal' option, which when selected produces a runtime

I love what you're doing here though! Please correct these issues, and generally double check your categories and this should be good

Very-Soft:
As an additional note, fingerless socks, and fingers and toes options appear inside of many different categories.

Poojawa:
As mentioned in the commit message, the duplicating markings were a result of the use of ^ operators. which is totally fine for the 'all' list as a filter, but it really wasn't sensible to do excessive for checks and check categories. so they've been yeeted entirely. Since we don't use all lists for everything anymore. 

the "Normal" option was a result of the get_available_styles() meant more for hair/tail/ears. since Normal = Null = base species features. Which... no species have base markings that get replaced!

I did shift both fingers and toes into the limbs category as well.

Very-Soft (changes requested):
It seems like the werewolf markings are available for people to pick exclusively in the 'all' category as the first few options, even while not using a werewolf species. These markings are intended for a very different sprite base, and it SEEMS like you correctly excluded them from the other lists, so I am assuming including them there for non-staff is unintended.

Poojawa:
As are the Vox and Teshari. 

There are no species checks at all, they have always been in that list, just completely buried with all the other ones.

Very-Soft (approved):
Hmh mh! I peek and test and that seems to be true. I'm still not the most happy about those showing up at the very top of the list, and I'd really prefer it if selecting a marking you already selected didn't reset the colors, but I can probably fix that part myself.

Poojawa:
They're simply listed as they arrive, since they're still in `Sprite_accessory_extras` they get 'found' first and added to the body marking. The solution would be to simply put them at the bottom of the body marking subfile

PR #211, opened 2024-04-10
Tubular

Very-Soft:


code/modules/recycling/disposal.dm
```diff
@@ -518,7 +518,7 @@
 
 			AM.forceMove(src.loc)
 			AM.pipe_eject(0)
-			if(!istype(AM,/mob/living/silicon/robot/drone)) //Poor drones kept smashing windows and taking system damage being fired out of disposals. ~Z
+			if(!istype(AM,/mob/living)) //Poor drones kept smashing windows and taking system damage being fired out of disposals. ~Z	// RS EDIT
```
Screemonster, line 521: will changing this break throwvore in the instances where a cargotech happens to be standing on the belt

Very-Soft: This line is for general disposal bins. In testing, people getting thrown randomly out of the bins results in a not-insignificant chance of getting thrown into a wall, which can cause serious injury.

The throw proc for disposaloutlet happens on 1587, and still includes people. However, frustratingly, in testing the flaps stop you from flying out of the tube, which is cringe, and I am working on.

Screemonster: cool and good (since people can't disposal-tag themselves like drones so they won't be popping out often anyway)

PR #233, opened 2024-04-18
Start to buff vore

lm40:
I made this a draft in case I missed anything obvious and also this hasn't been tested at all yet but the general idea is to make vore more filling, tho I also doubled the stuff going on with healbelly since more nutrition should be common anyway


lm40:
Ok I have something like tripled most vore nutrition gain but also adjusted the nutrition displays so they are a little bit more organized

Very-Soft (approved):

PR #243, opened 2024-04-25
Couple new verbs

Very-Soft:
Adds Leap, which is a generic proc for any mob/living. This verb can be applied via fun > toggle event verb

Also adds a verb to corrupt hounds to toggle their step sound.

### DIFF
--- a/code/modules/eventkit/RS-event.dm
+++ b/code/modules/eventkit/RS-event.dm
@@ -83,7 +83,7 @@
 	if(!choice)
 		return
 
-	visible_message(span("warning","\The [src]'s eyes flash ominously!"))
+	visible_message(span("warning","\The [src]tenses up in preparation to leap!"))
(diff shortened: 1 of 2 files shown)


code/modules/eventkit/RS-event.dm
```diff
@@ -56,3 +58,50 @@
...
+
+	if(!targets.len)
+		to_chat(src, span("warning","There are no valid targets in range."))
+
+	var/mob/living/choice = tgui_input_list(src, "Pick a target! (Only those with viable mechanical prefs are included)", "Target Choice", targets)
+
+	if(!choice)
+		return
+
+	visible_message(span("warning","\The [src]'s eyes flash ominously!"))
```
foopwo, line 86: Should probably have a different message, not everything has (visible) eyes!

PR #253, opened 2024-04-30
Hypospray MK.IIs

Poojawa:
Ported this wholesale from Citadel, my code, adjusted for this code base.

Hypospray and Mini-kit sprites by CameronWoof, hypovials by me.

bb6b56a5425cf8d713c8117561b0d27c

Adds ergonomic Hypospray mk IIs. Currently only available in the NanoMed Plus pending a balance discussion. They were also enabled to be placed on labcoats and medical belts, including the alien one because it was there.

The Hypospray MKII is capable of both spraying and injecting whatever reagents are loaded in the hypovials. They only accept hypovials, which only have 30 and 60 units of capacity for small/large. They follow the same rules as autoinjectors other than having a chemistry-friendly mixing ability and application mode. The hypovials control the injection amount, the hypospray is just the delivery system. 

these hyposprays are also Alt/Ctrl click context enabled, so you can quickly swap from inject to spray, or quickly eject a vial if your hands are full.

Hypovials cannot be drunk from. I want to make them act like sealed containers without being sealed containers, but the interactions just aren't in the code. 

 Using a pen on the vial will allow you to label or change the visuals of the vial itself for preference. There are bluespace versions, which simply double the amount the vials can hold, and have an overlay that persists through recolors. They haven't been added to science, again pending balance discussions.

Currently the CMO and Combat versions, which are double the size and faster, are unobtainable, but if ERT needs 'em I'm sure the admin handling that can spawn 'em. There's pre-packaged boxes that can be spawned. Boxes are only capable of holding the mkii hyposprays and hypovials, and you can only stuff 7 into 'em. No free analyzers either. 

Reagent Containers also got a buff, in that Alt clicking now brings up the transfer amount dialogue.


Very-Soft:
Don't know enough about this and too eepy to do much learning, so it will have to wait for right now, will try to review and test after eep

code/modules/reagents/reagent_containers/hypospraymkii.dm
```diff
@@ -0,0 +1,482 @@
...
+	add_attack_logs(user, L, "[user] attemped to use [src] [L] which had [contained]")
+
+	if(!do_mob(user, L, inject_wait))
+		return
+
+	if(!vial.reagents.total_volume)
+		return
+	add_attack_logs(user, L, "[user] applied [src] to [L], which had [contained] (INTENT: [uppertext(user.a_intent)]) (MODE: [fp_verb])")
+	if(L != user)
+		L.visible_message("<span class='danger'>[user] [mode ? "Injects" : "Sprays"] with the [src] on [L]!</span>", \
```
Very-Soft, line 209: The grammar here could be improved. It's also a little weird that on 195 you gather which verb to use, but here and on 210 you use get it again and capitalize it.

would probably be better like '\The [user] [fp_verb]s \the [L] with \the [src]!

code/game/objects/items/weapons/storage/firstaid.dm
```diff
@@ -167,6 +167,67 @@
...
+	name = "hypospray mk.II kit"
+	desc = "Contains a Mk II Hypospray and its hypovials. It can only hold mk.II related items."
+	icon = 'icons/obj/storage.dmi'
+	icon_state = "firstaid-mini"
+	starts_with = list(/obj/item/weapon/hypospray_mkii, /obj/item/weapon/reagent_containers/glass/bottle/hypovial/small = 3)
+	can_hold = list(/obj/item/weapon/hypospray_mkii, /obj/item/weapon/reagent_containers/glass/bottle/hypovial)
+	//can fit large and small vials, and any of the mkiis. but that's it. No free Analyzer either.
+
+/obj/item/weapon/storage/firstaid/hypokit/tricord
+	desc = "Contains a Mk II Hypospray and its hypovials. It can only hold mk.II related items. This one is prefilled with Tricordazine."
```
Very-Soft, line 185: It would be good if the descriptions on the kits and the vials mentioned what problem they help resolve. Will help newer players figure out what does what.

Very-Soft:
The other feedback from the medical player was, that they weren't sure that spray actually does anything. They saw that it applies medicine to the skin, but they weren't sure that that actually does anything with our medicines, so weren't sure how useful the spraying would be.

They also mentioned that tricord and oxy together was pretty strong, and that the oxy would make vision blurry for a while. They didn't say it should be changed, but just mentioned it as something to look out for.

code/game/objects/items/weapons/storage/firstaid.dm
Poojawa: the kits are color coded, for a visual ID

Poojawa:
The lack of practical applications is a chemistry issue, then. out of scope for this PR. adding topical layer chems would, naturally, be of great use. Tier 2 chemicals maybe. 

the CMO's pre-mixes are super strong, but I just copied the ingredients for the combat autoinjectors, sooo.

code/game/objects/items/weapons/storage/firstaid.dm
Very-Soft: I understand that they are color coded, a new player will not know that color code though.

Poojawa: Then that new player should not be interacting with potentially dangerous medical supplies on their own. 

Or they'll learn by doing. Either way, we shouldn't be excessive with the swimming floaties over hypothetical what-ifs. A truly helpless new player like you're suggesting is a failure of more experienced players not teaching them.

PR #270, opened 2024-05-05
Buffs solars and tcomms power

Very-Soft:
Doubles the number of solar panels through the magic of multiz and also buffs the cells in the telecomms apcs


maps/stellar_delight/stellar_delight3.dmm
```diff
@@ -12954,8 +13160,10 @@
 	icon_state = "0-2"
 	},
 /obj/machinery/power/smes/buildable{
-	RCon_tag = "Substation - AI/Telecomms";
```
Poojawa, line 12957: regression?

PR #277, opened 2024-05-06
Tool

Very-Soft:
Adds a tool for staff to use to select the next round's redgate and/or gateway map. Only affects the next round, not the current round or any rounds after the next round.

Also makes GB not have only fantasy castle.

### DIFF
--- a/code/game/Rogue Star/special_persist.dm
+++ b/code/game/Rogue Star/special_persist.dm
@@ -6,9 +6,10 @@
 /datum/persistent/next_round_maps
 	name = "next round maps"
 
-/datum/persistent/next_round_maps/Shutdown()
-	var/filename = "data/persistent/next_round_maps.json"	//Let's just hard code this I'm sure it will be fine
+/datum/persistent/next_round_maps/SetFilename()
+	filename = ADMIN_CUSTOM_MAP_LOAD_PATH
 
+/datum/persistent/next_round_maps/Shutdown()
 	var/list/to_save = list()
 
 	if(SSpersistence.redgate)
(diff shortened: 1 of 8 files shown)


code/game/Rogue Star/special_persist.dm
```diff
@@ -0,0 +1,77 @@
...
+//If you load one map with one set of redgates, and another map with a different set, it won't matter what you saved
+//you will just have told the system to load something that may or may not exist in the code at that time
+//so keep the lists of anything we save here synced between all maps or it won't work right!
+
+/datum/persistent/next_round_maps
+	name = "next round maps"
+
+/datum/persistent/next_round_maps/Shutdown()
+	var/filename = "data/persistent/next_round_maps.json"	//Let's just hard code this I'm sure it will be fine
+
```
Arokha, line 11: I see that SSpersistence is kind of bad and annoying. I think you should still use /datum/persistent/SetFilename() and have that proc do one thing only, set the `filename` var (a var on /datum/persistent which already exists, rather than this local one) to your filename, and then read that when deleting/saving it etc down below (see paintings.dm).

I kinda wanna just rewrite SSpersistence and I'll probably make that var more meaningful, so while you're HERE and have this OPEN maybe just do that.

code/game/Rogue Star/special_persist.dm
```diff
@@ -0,0 +1,77 @@
...
+	if(SSpersistence.redgate)
+		to_save["Redgate"] = SSpersistence.redgate	//We're saving a string for the kind of thing, and a string for the specific selection as an associated list
+	if(SSpersistence.gateway)						//Your map list needs to have the name you want saved as the index for the actual map names to be compatible
+		to_save["Gateway"] = SSpersistence.gateway
+	if(fexists(filename))	//Delete the file so we can save a new one
+		fdel(filename)
+	if(!to_save.len)		//Nothing to save, let's just quit here
+		log_debug("No custom next round data detected, aborting Shutdown function.")
+		return
```
Arokha, line 22: It might be nice to log the difference of 'I removed the existing one and I'm not replacing it' vs 'There was no existing one and I'm not making a new one'. Maybe just adding a log message for when it removes an existing one would suffice.

Arokha (approved):
If you test most recent commit and it works then I think it is gud
I still wanna rewrite SSpersistence now though.

code/game/Rogue Star/special_persist.dm
```diff
@@ -0,0 +1,79 @@
+//PLEASE NOTE, this system will only work correctly if the affected map lists are the same on every map load.
+//If you load one map with one set of redgates, and another map with a different set, it won't matter what you saved
+//you will just have told the system to load something that may or may not exist in the code at that time
+//so keep the lists of anything we save here synced between all maps or it won't work right!
```
Poojawa, line 4: This would probably be good advice as well for future main maps. More universal, generic mapping areas so you can flip through the primary maps without having to *completely* recompile just to swap from SD to RP in the map editor.

PR #283, opened 2024-05-07
Ports ribbons from chompstation

Aphast:
as it says, just adds ribbons from chompstation as a marking
image
pictured here


Arokha (approved):

Aphast:
Small edit, it turns out that the ribbons from chomp are actually backwards and I only just now noticed even after months of firsthand experience, this fixes the issue by swapping the forward and backward states, character changed to a solid two-tone to emphasize contrast that the 'bow' is on the front of the neck, not the back
image

Very-Soft (approved):

PR #285, opened 2024-05-09
Ghost Mob Misc Buffs

Very-Soft:
Makes it so that ghost spawn mobs start with a radio to listen to

Also restricts mob radios so that they can only be spoken to if you have galactic common set to the default language, to prevent things that can't speak common from using the radio.

The idea here is to give players the tool so that they can be informed about what's going on on the station, not necessarily to let them talk. I figure, if a mob can speak common, it can probably logically use the radio.

Most animal mobs cannot speak common, so this shouldn't let things like dogs and cats yell their animal sounds on the radio, while still allowing them to hear what's going on.

Adds all the oties to the ghost mob selection, organizes the list a bit, enables the relevant procs to work with nested lists for organized sub selections.

Also adds thermal vision to maint preds and ghost join mobs to help enable them to do spooky predator things

### DIFF
--- a/code/modules/mob/ghost_simplemob_spawn.dm
+++ b/code/modules/mob/ghost_simplemob_spawn.dm
@@ -20,19 +20,34 @@ GLOBAL_LIST_INIT(ghost_spawnable_mobs,list(
 	"Frog - Giant" = /mob/living/simple_mob/vore/aggressive/frog,
 	"Jelly Blob" = /mob/living/simple_mob/vore/jelly,
 	"Juvenile Solargrub" = /mob/living/simple_mob/vore/solargrub,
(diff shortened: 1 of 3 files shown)


code/modules/mob/ghost_simplemob_spawn.dm
Malfaren, line 1: Only thing I'd ask since we're getting sec-otie in would be two other additional variants for flavor. The red and phoron sec oties.

"Otie - Guard Red" = /mob/living/simple_mob/vore/otie/security/phoron/red,
"Otie - Guard Mutated" = /mob/living/simple_mob/vore/otie/security/phoron

Malfaren (approved):

PR #287, opened 2024-05-09
Small buff update Talon crew

AzzyDreemurr13:
## Why and what will this PR improve

As an active Talon player, I noticed and learned that unfortunately nerfs was very strong from VStation and now all expeditions are extremely difficult and problematic (in other words, they have become very stuffy and now all players often explore Red Gate, and the expedition department was mercilessly destroyed). I would like to start a series of updates for Talon and a couple of small changes for a couple more Talon Jobs.

### It was also added to the Talon Cap Locker:
- Helmet 
image
- AR-B glasses (For some reason, they haven't been around for a long time lol, because in Vending machines on Talon you can get another kind of AR Glasses, so it's strange that the Captain doesn't have their cool AR glasses)
- Command insignia
image
- Talon Officer Beret
image
image

#### It was also added to the Talon Guard Locker:
- Helmet 
image
- Holster 
- The extremely weak and old type of armor has been replaced with a regular armor with pockets like Station Officers
image

#### It was also added to the Talon Engineer Locker:
- Engineering insignia
- Meson scanner HUD (I added this because it is not used anywhere, the sprite is interesting, and ordinary Station engineers have meson glasses according to the standard. So now the Talon Engineer will be able to boast of interesting equipment)
image

#### It was also added to the Talon Lockers:
-  insignia related to their job
- Talon Caps 
image
- Talon Bar
image

## Changelog
:cl: Azzy.Dreemurr
- a locker with musical instruments added in Talon Bar
- Talon armory got more holsters itself
- Talon Captain got helmet and Talon officer beret
- Talon guard got holster and armor upgrade
- Talon Engineer got Special Meson HUD Glasses
- All job Talon Lockers now have the insignia of their department, not just 1 Guard Talon


FartMaster69420:
Right now Talon is sitting in a similar spot that exploration was in when that was a thing. The problem being an over reliance on combat for their content giving a murder hobo reputation. This update is just going to make the combat easier with no real additions to RP at best.

Further, gatekeeping the new gun to the Captain's locker feels like a push toward Captain being more of a main character of Talon. I feel like making the captain the best soldier on the ship kind of misses the point of what the role should be doing, which is facilitating RP for the rest of the crew.

As for the change to the access for the captain's locker I have to question if it would be needed should the gun not be added. As it stands, the worst that could happen is someone take a fancy voidsuit, hat, or armor vest.

As a whole, I don't feel like giving Talon an extra gun or two is a BAD change, but should be considered carefully and probably should come much later after OM content is reworked and Talon has a better reputation.

Michab02:
I don't think it's a good idea to give the captain a fancy assault rifle. They're the captain, not a swat team. At most I'd maybe put one or two riot shotguns in the talon's armory, imo...

FartMaster69420:
I’d say not to replace the guard’s armor since it actually has a unique ability compared to the sec armor.

the plate carrier can have its armor plate replaced with a plate made from any material which depending on the material can make it a lot better than a standard vest.

crossexonar:
So after playing talon for roughl a week, both with Station Volunteers and without, I can firmly say that other than the flavor additions, most of this just, isnt needed. New insignias Cool, Engineer hud, Cool, They don't need any additional weaponry as it can be found Through Poi's or purchased through Station (provided the SM/HOP doesnt run you through paperwork for an hour)
So, Feel free to add flavor, but don't change the balance.

Michab02:
A midtier weapon in the armory is a lot more reasonable imo, so I'm not against this. Talon has been kicking off with RP lately too.

lm40:
I don't really think Talon have any need of more guns ATM, but the other changes are mostly fluffy so I don't much mind them.

lm40 (changes requested):
I am fine with the insignias, the engineer changes, the officer changes, and probably even the holsters but I think the gun is unnecessary

lm40 (approved):

PR #288, opened 2024-05-09
Rogue Star not Vorestation

Poojawa:
Honk. I'm sure this won't break anything at all ever. Right?

I'm sure there's dozens more settings to change so maintainers pls edit database relevant files

rust_g probably needs a fork/recompiling, but I do not have access to that, so it's still default.


lm40:
I am not really interested in having this, it seems likely to cause problems and brings zero tangible benefit.

PR #289, opened 2024-05-09
Ports dual-color shadekin ears

Aphast:
Per request, dual color shadekin ears from chompstation stored in the ears 64 file, since this base does not have ears-ch. high-contrast character for easy visibility:
image
:cl:
add:Dual color shadekin ears
/:cl:


code/modules/mob/new_player/sprite_accessories_ear_snowflake.dm
```diff
@@ -137,6 +137,15 @@
...
 	)
 
+/datum/sprite_accessory/ears/shadekin/dualcolor
+	name = "Shadekin Ears (Dual Colorable)"
+	desc = ""
+	icon = 'icons/mob/vore/ears_32x64.dmi'
+	icon_state = "shadekindual"
+	do_colouration = 1
+	color_blend_mode = ICON_MULTIPLY
+	extra_overlay = "shadekin-under"
```
Poojawa, line 147:
```suggestion
	name = "Shadekin Ears (Dual Colorable)"
	icon = 'icons/mob/vore/ears_32x64.dmi'
	icon_state = "shadekindual"
	extra_overlay = "shadekin-under"
```

No need for the description add and tbh I forgot to remove do_colourization and color_blend_mode for the rest of these since both of these entries are default now

Aphast:
makes sense, I would have changed more of the values but the confusion of trying to find the default shadekin ears in the first place caused more than enough issues. I genuinely don't know where the default skin icon files are so I decided to just directly port from chomp with no real alteration to see if it worked. From the player/testing perspective it appeared to so *shrug. Regardless, I more than trust such changes.

Poojawa:
Yup yup! Totally understandable.

trying to compress the amount of icon file spread for exactly that reason too, since it's so completely a headache to deal with, haha.

PR #293, opened 2024-05-11
Un-dogs the non-dog borgs

foopwo:
### **THIS PR SHOULD HAVE NO PERCEIVABLE EFFECT FOR DOG/DRAKE/MEKA/WHATEVERBORGS, ASIDE FROM THE COMBAT BORG FIX**

Reskins all dog-related items for non-dog borgs to have nothing to do with animals. Everything is functionally the same item but with icon and text changes.

boop module becomes a generic scanner
pounce is now the R.A.M. module
the tongue is now the Mrs. Fusion portable reactor
the dog bellies are now more generic bellies, I'm not a good spriter, but I would rather see those than doggy ones

Also fixes not being able to equip a couple combat borg modules! Wow!

borgbellies

Let it be known here that I absolutely loathe the roundabout manner that this is done and I would've preferred regular borgs not getting the dog items in the first place.


foopwo:
PR is finished.

PR #307, opened 2024-05-14
Rascal's Pass Engineering Rework

Michab02:
image

This is a massive rework of Rascal's Pass engineering. There is now a showers/bathroom and a cozy break room with a coffee machine for the engineers, all branching off of a central hallway. The locker room has been sorted and organized, windows added, and the CE's office placed centrally. (But also shrunk a little) There's now a larger engineering warehouse holding all the bits and bobs, as well as the voidsuits, and looking snazzy. And finally, maintenance has been expanded, with multiple ways to get into the caves and and an _underground lake_ to explore!

There may (likely) be issues with this due to the scope, so please let me know here or on discord if you notice anything and I can fix it up right quick!

image
image
image


Michab02:
Also, I did stress-test the atmos to the best of my ability with a phoron fire and it worked as intended~

PR #308, opened 2024-05-14
[READY] Power overhaul, Layered Cabling

Poojawa:
This is an attempt to implement several things for mapping and quite possibly in game QOL

Layers!
944f52f04730b79088d0640a44c02f56
 Clicking wires in hands!
321cc8a90e09d733b2c14f6a380d8242

SMES Panel!
4653260cc16f0e693f0ab031bbe98bd5

Replacement for ender  cables! Power transmitters. Function the same but they aren't just a heavy cable node
bc4edd7991483b773c7848f44c0410dd
14cd55effd3a7bdf996ceb5aa37f123a

Sprites credits: 
Power transmitters done by me.  
Cables are just offset base ones. 
SMES units were already present, I added the panel open stages. 
Wire bridge/relays from /tg/, I added relay's on/off states
RCL from /tg/ although I'm responsible for their color overlay fix via Citadel
Pipe cleaners are just old cable icons
Major Changes:
- Cables are now able to connect to each other, assuming the same layer, without needing the user to pixel click the wire sprite. 
- Cables only connect on cardinal directions. Diagonals no longer exist due to sprite issues and overall usability.
- Cables have in hands interactions that allow for swapping layer, using a multitool after will change the color whilst preserving the layer chosen. This resets if the layer is changed.
- Heavy Duty cables were converted to use the 'smart' system as well, operating on its own layer entirely. Limited amounts on the map are intended. They also will not interact with most everything else that regular cable coils do like repairs or construction.
- Cable bridges, constructed via inhands radial of a cable coil, allow engineers to merge powernets of various layers into a singular one. clicking with an open hand to open the selection radial
- Cable bridges are deconstructed simply by using wirecutters on them. 
- Deck relays replace the up and down cable structure ends between Z-levels. They are also constructed via inhands usage of cable coils. They can be cut with wirecutters or unanchored with a wrench.
- Deck Relays only utilize a single powernet layer, it proved too difficult to enable them to maintain multiple layers and thus, multiple powernets at the same time. Perhaps this can be revisited.
- Power transmitters have replaced 'Ender' cables in long distance/Z-level connections. As a full machine they are more easily utilized than the old cables. Set via multitool for ID. They are only able to connect with heavy duty cables under them.
- Breaker Boxes are utilized as power switches again. Their RCON button has been updated with 'Isolate' rather than 'Bypass'. More information with substation changes. You can also wrench and move them if needed. Their interactions have been considerably sped up as well.
- SMES Units have been standardized for construction. Their icons have been updated with sprites formerly only used by the 'hybrid' SMES units, and are almost identical to current Baystation sprites.
- SMES units have preserved the ability to attach more than one terminal at the same time to bolster their charging, although in practice this was never actually used because no-one knew about it.
- SMES units can be saved from a overload state by attacking it with a metal rod. Preserving the unit but also launching a tesla bolt with all the SMES' charge out. They previously were unrecoverable and would explode without any means to stop it.
- SMES Units provide admins with readable investigate logs under 'powernet', showing settings and charge.
- SMES Units have been given mapping-assistance with preset versions, in order to reduce the amount of inconsistent map var editing.
- the Engine Core SMES unit is meant to be the sole reciever of engine power, in order to prioritze containment power. The Mains SMES starts input toggled off, in order to prevent draining the Engine core of energy should there be a delay in engine setup. 
- Substations have been remodeled to ensure that they are the primary provider to department power output. The breaker box toggle now can isolate the subdepartment from the main grid, for use in powering shields or PD systems. A cable bridge can be built in the room to direct wire the department to the main grid to bypass the insulation effect. 
- Substations being set up this way means powernet sensors are actually readable and useful, only displaying the department APCs as intended.
- Emitter code was refractored to improve backend performance. 
- Field Generator code was refractored to improve reliability of staying on by dividing all its power expectations by a literal thousand. Now they recharge simply off energy damage, rather than a complex conversion into kilowatts.
- Field Generators no longer connect if any dense atom is between them. This was an unrealized oversight fix as the code was 12 years old when it was last touched. The code was not updated upon introduction to Tesla engines, and previously only closed airlocks prevented connections, as it was assumed anything else would have been destroyed by the Singularity's normal operations. 
- This change can be partially reverted with a check to exclude power machinery (Coils, Emitters, grounding rods, SMES units, etc) from preventing the connections. 
- Rascal's Pass and Stellar Delight were given complete wiring overhauls. Pathing of cables was redirected through doorways whenever possible. All departments are now on a seperate layer, either Layer 1 or 3, Layer 2 (being the default) is regulated to be the Mains specific that the SMES outputs to. As a reminder, cable bridges allow for on-the-spot overrides if a bypass is required.
- Solars were cleaned up to just the cable under them, the wire breaker map tool still works to randomly disrupt connections.
- With the retirment of old connection behavior and its inpact on possible wire art, 'pipe cleaners' were introduced to preserve that functionality in its entirety. In addition, pipe cleaners have been enabled to be placable on carpets, tile, grass, and more. Public displays are now possible without tearing up the floors.
- Grilles will auto-connect to shock if there is a wire under them, placing a tile on top disables their shock ability. This change was made due to usability concerns.
- Powernet propagation code has been overhauled, it should work faster and not cause a hiccup when someone trims cables and requires a rebuild. 

Minor changes:
- multitool usage on wires now displays Available power like before, but also Demand and Surplus of power. 
- Multitools have interactions with various machinery to hard set their connected layer. If the multitool is used for something else (Smes RCON, Power tap ID) then using a cable coil will bring up the same dialoge to change the layer.
- Cable coils are assigned the has_tool_quality(TOOL_CABLECOIL) functionality, in line with other tools. This will also be used for the RCL when it can be implimented. 
- Cable Layer machinery and Mech module have been removed in preperation for handheld and borg mounted Rapid Cable Layers to assume their role. They were never used in any capacity anyway.
- Debug mapping verbs have been enhanced to display more accurate locations, as well as enabling a 'jump to' button for quicker fixing in SDMM. 
- All maps, Pois, gateway, redspace, Overmap ships were given a clean sweep in SDMM. They are technically all smaller in file size as they've had at least a third of their vars cleaned up.
- Rascal's Pass and Talon had their atmosias upgraded to have connected gas meters to all relevant portions of their atmospheric systems. Gas monitor terminals were placed in monitoring areas and their bridges.
- Talon's nuclear reactor now visually glows when overheating, the sprites were in the file, but no-one had enabled them.
- RTGs were shuffled out of port_gen_vr and into their own file for readability and consolidated. 
- APC_vr was transfered into the main APC file. 
- Emitter_vr was rolled into Emitter for the most part. 
- SMES_construction and SMES_vr subfiles were rolled into SMES_subtypes to reflect their consolidation
- cable_logic was deleted, due to having been unchecked and unused for > 10 years. Mostly housekeeping reasons.
- cable_ender was deleted, its functionality was incompatible with initialization of cables and was replaced by power_transmitter
- Power related icons were mostly moved from scattered positions in the icon folder to icon/obj/machines/power and related subfolders.
- power define file created for maintainability.
- Cable layer colors can be set via the defines in the file, any changes will update all cables and radial icons to the new color.
- Cable coil inhands were given greyscale sprites in order to dynamically update visually
- Terminals reflect the connected layer via colorizing their wires.
- various power related is_helpers added. 
- DisplayPower updated to show 0Ws instead of otherwise confusing negatives for Surplus amounts. 
- DisplayPower macro utilized more heavily for Admin logging readability. 
- Machines/Cells/etc had their power draw numbers converted to KILOWATTS, MEGAWATTS, or WATTS macros, for readability in the code. 
- Electrocute Mob effect had the damage powerscale from Virgo/Chomp ported in and optimized.
- General maintenence on attackby interactions to use modern code such as default_unfasten_wrench or has_tool_quality, same with replacing opened with panel_open 
- cleaned up old font code and spans with span_<level> macros, for readability.


FartMaster69420:
So does this completely remove the ability to build wire art into your engine constructions?

Poojawa:
> So does this completely remove the ability to build wire art into your engine constructions?

Yes. 

Pipe cleaners exist to provide wire art without powernet lag.

Poojawa:
```
[2024-05-22T14:47:31] Runtime in cable_heavyduty.dm,112: type mismatch: 0 |= the large power transfer node (/obj/structure/cable/heavyduty/ender)
  proc name: Connect cable (/obj/structure/cable/heavyduty/ender/Connect_cable)
  src: the large power transfer node (/obj/structure/cable/heavyduty/ender)
  src.loc: the plating (66,71,4) (/turf/simulated/floor/plating)
  call stack:
  the large power transfer node (/obj/structure/cable/heavyduty/ender): Connect cable(0)
  the large power transfer node (/obj/structure/cable/heavyduty/ender): Initialize(1, null, null)
  Atoms (/datum/controller/subsystem/atoms): InitAtom(the large power transfer node (/obj/structure/cable/heavyduty/ender), /list (/list))
  Atoms (/datum/controller/subsystem/atoms): InitializeAtoms(null)
  Atoms (/datum/controller/subsystem/atoms): Initialize(532488)
  Master (/datum/controller/master): Initialize(10, 0, 1)
```

```
/obj/structure/cable/heavyduty/ender/Connect_cable(var/powernetless_only = FALSE)
	. = ..() // Do the normal stuff
	if(id)
		for(var/obj/structure/cable/heavyduty/ender/target in cable_list)
			if(target.id == id)
				if (!powernetless_only || !target.powernet)
					. |= target
```

sigh

FartMaster69420:
Can't say I'm a fan of this forcing all wiring to be grid based, removes a lot of flexibility in wiring beyond just wire art.

foopwo:
Layers adds a lot of flexibility that was hard or impossible to do before. Though I will miss making pretty (or extremely cursed) curvy wires.

Poojawa:
> Can't say I'm a fan of this forcing all wiring to be grid based, removes a lot of flexibility in wiring beyond just wire art.

"wiring to be grid based"

It...it already is ?? the only thing this makes obsolete is diagonal connections which were scuffed and barely functional to begin with

and there's three entire layers of cabling now. which do not mix at all unless connected together by the Bridger. I don't understand how much more flexibility you even need for wiring projects beyond that.

Poojawa:
Will fix merge conflicts when Autumn is done fiddling with RP.

But I'm pretty sure this is good to go.

Poojawa:
All the maps were either edited to have their power cables changed, or simply given a clean up in SDMM via opening -> saving. only the primary maps and the common loaded sub maps (aerostat, overmap ships) were touched to ensure they look nice and had sane cable layouts.

SD got multiple pipe fixes that were just simple mapping overlooks, like a manifold instead of a corner piece, a length of pipe that didn't match the others, that sort of thing. I also ensure SD's brig cell A is actually the 'A' block, not just the base, as well as the AI/Tcomm area is a dedicated Substation area

Poojawa:
When the unit tests just straight up runtime without any trace. It's great!

code/modules/power/smes_construction.dm
```diff
@@ -1,65 +1,137 @@
...
+	output_level = 1 MEGAWATTS
+
+// Mains gets a little extra snowflake to help early rounds when no engineers are around, mostly for outputting
+// This is about the same amount with previous dirty var editing.
+/obj/machinery/power/smes/preset/mains/Initialize()
+	. = ..()
+	component_parts += new /obj/item/weapon/smes_coil/super_io(src)
+	component_parts += new /obj/item/weapon/smes_coil(src)
+	component_parts += new /obj/item/weapon/smes_coil(src)
+	recalc_coils()
```
Poojawa, line 71: I wanted to provide the same gist of power boost that was dirty map edited, but more consistently. Since the Engine core is separated off and the Mains doesn't input, Engine power should last even if the station goes dark. 

It's just a matter of tweaking starting power numbers at this point

FartMaster69420 (changes requested):
Restricting wires to cardinal directions only and enforcing smart wiring is a no-go for me.

When I play engineering I like to make custom designs when wiring the engine, this adds the option to add an artistic flair to what is otherwise a mundane setup. The smart wires just removes that entire dynamic with nothing to take it's place.

I suggest leaving old wires as their own layer to preserve the functionality.

Poojawa:
> Restricting wires to cardinal directions only and enforcing smart wiring is a no-go for me.
> 
> When I play engineering I like to make custom designs when wiring the engine, this adds the option to add an artistic flair to what is otherwise a mundane setup. The smart wires just removes that entire dynamic with nothing to take it's place.
> 
> I suggest leaving old wires as their own layer to preserve the functionality.

uh, not possible to implement diagonals. it would result in complete, horrid messes that would look terrible. previous wire art also did horrible things to powernet propagation pathfinding, causing immense lag on cutting/placing cables. This new system is better suited. 

If you want to make pretty wire art, pipe cleaners exist. They can be placed on plating or full tiles as well. they operate exactly as you're used to, and can be colored in hands with a multitool with no additional cost. 

Your suggestion is otherwise unable to be accommodated.

code/modules/power/smes_construction.dm
Poojawa: gave Engine core a Super_IO and Capacitor, which gives a decent enough amount of flexibility for basic engine set up.
Mains have a decent enough mix as per my quick testing

SD's PA might be a lot more finicky due to being on the subsystem grid, but that's good as well.

Michab02:
There are a few weird issues with this right now as of testing - emitters sometimes spawn anchored or not connected to cables in spite of emitter code not being touched. Or already welded but not anchored, which is very strange.

The default SMES arrangement is also liable to cause a tesloose because the main SMES will suck all the power from the engine core SMES and deprive the emitters of power, if both SMES are set to max. I strongly suggest changing this, since a lot of engineers do that based on muscle memory and it can cause an easy tesloose. The current set up of SMES on live is more ideal, imo:

image

Poojawa:
As per our discussions in DMs:

Resolved the emitters self-setting themselves to the wrong state. It needed a TRUE/FALSE in should_have_node(). TG already has welded as a machine level var, so I just threw it into emitter code as a holdover. 

Adjusted Engine Core's maximum baseline in/out, in order to maintain engine output and power for the mains. It can still be upgraded on its own, but this should allow for maxmimum usage of absolute base tier power setups. Engineers are to be reminded that Mains must not exceed the power draw of Engine core's available output. But engines getting messed up because someone isn't paying attention is a feature, not a bug. 

We also resolved the topic of the substation's alternative power layout, Bridges continue to exist to achieve the previous functionality of direct smes bypass if an engineer wishes to do so. but a substation is now directly in control of powering (and protecting the main grid from outside power drains like solargrubs). It was also mentioned that isolating the subdepartments now has an appreciable effect on maximizing main grid power in an emergency (Shields, PDS) via RCON.

Zoldorf:
"Cables are now 'Smart' in that they will auto connect on cardinal directions without having to click the knot wire and hope for the best. They will also update the color in your inhands sprites!"

This is a no, please, do NOT do this, this is actually miserable to people who actually know and play engineering.

I would much rather see the "knot" hitbox just be made invisibly larger like how the coils work to make it easier to click instead of having these dreadful "smart" (read: dumb) wires that "do it for you" when in reality, all it does it make actual good and clever wiring impossible, not to mention wire art becoming entirely extinct.

I will put together a list of things that become impossible:

Wire-art
Wiring diagonals (If you don't know why this is important, why are you even changing this?)
Running multiple wires under machines but only connecting to 1
Close proximity wires that are connecting machines and are on the same layer
Terrible wire gore (It's funny)
Running wire under grilles without making them electrified
Running past anything without connecting it

Most of these things are actually important for when you are doing hands on work on the server itself in character.

Poojawa:
Implying I am not an engineering main...?

> Wiring diagonals (If you don't know why this is important, why are you even changing this?)

It used to be important because you used to have to be conservative with your cable coils, since it took a minimum of two to place a full connection per tile. Now it merely takes a single length so you effectively have twice as much coil as before. Diagonals aren't actually needed anymore from a mechanical standpoint.

> Wire art 

again, Wire art is terrible for server performance, and is so infrequent a thing that happens that it's almost entirely 'once a month at best' type of ordeal. I'm still not convinced that such terrible downsides should be maintained for a single gag that doesn't get seen by anyone and isn't done very often at all.  If you wanna make stuff pretty, use the pipe cleaners. It's the same. Exact. code. as old cables but without any powernet interactions. I don't know why I have to keep repeating that point. 

> Running multiple wires under machines but only connecting to 1

running connections directly under machines that do directly connect are so few that this isn't really an issue either, as only shield generators and grilles have direct connections that are actually used. Did you not already know this? Not even the PA is direct wired anymore, it just runs off the APC.

> Running wire under grilles without making them electrified

It would, however, be reasonable to change grilles to not auto connect if the floor is tiled, but I'd like staff input on that aspect. 

> Running past anything without connecting it

You also miss the point about layers. They do not connect between each other, so if L2 is arcing off to connect to something, L1 won't. you have **three** layers of cable to build around. Ways to merge those layers into the same powernet, and easier overall management of said wires.

Michab02:
Shield generators, emitters, RCAs, pacmans, solars, and Tesla coils all connect directly to wires as well, and some of those get set up every shift.

Poojawa:
> Shield generators, emitters, RCAs, pacmans, solars, and Tesla coils all connect directly to wires as well, and some of those get set up every shift.

power machines are supposed to connect to a node, yes. That is a given. however if they're already connected to a layer, in most cases they will not reconnect to a different node unless fiddled with. deck relays periodically recheck connections since powernets sometimes just need that. 

I was experimenting with ensuring said power machines would only connect to a certain layer, although we don't have as many interaction abilities as tg does, as they have a dedicated right click attack check they put that interaction easily on. 

but unless someone is deliberately trying to ultra micro build - and even then if you don't give yourself physical space to move around there wouldn't be any conflict to using a different layer. the bridges exist to mix powernets for that edgecase.

FartMaster69420:
Seems like a huge step backwards to go from being able to wire in 8 directions to only being able to wire in four.

The trade off of more efficient use of cable coils is negligible at best, considering how many of those spawn around the map. Even in my most inefficient setups I've never had issues running out of coils.

Poojawa:
Try it for yourself then.

BriarGarden:
As someone who's done a lot of engineering myself, I don't see any issues here. Building random shit when I'm bored is one of my go-tos, and I've literally never encountered a situation where I _need_ diagonal wires. Further, until this PR was put up, I'd seen wire art _twice._ In those cases it wasn't even art, someone just...made the wires squiggly heading towards emitters and collectors. It's only now that the PR has come up that I've suddenly seen people do it more often and complain that it's being taken away and even then. Still just a couple squiggly wires at the engine, and used....literally no where else.

In short, I'm basically saying this is making a mountain out of a mole hill at best. If you want to make colorful designs on the tile somewhere, the pipe cleaners exist to facilitate that just as well as the wires did. Likewise, since a majority of machines connect via APC, you can just run pipe cleaners for that cabled up look you want to do. It'd be just as useful mechanically as using wires currently for that task, because they don't interact with them anyway.

The only thing I can say is decently valid complaint, is electrified grille. Since it would be a bit of a pain to work around if you don't know it'll electrify no matter what, but on the same note I don't know why you'd run your cables under grilles outside the bridge. Common building techniques usually avoid building under walls/grilles for ease of modification, usually running through a doorway so any breakages can be easily mended or splits can be easily done. 

All other complaints stem from a time when everything wasn't loaded to APCs. As was noted earlier, the only machines that direct connect deal with power. IE: Shield generators, emitters, RCAs, pacmans, solars, tesla coils, rad collectors, etc. Not even the particle accelerator needs to be directly wired anymore.

FartMaster69420:
> Try it for yourself then.

???

FartMaster69420:
Okay, after some experimenting on the test server I have some thoughts.

I like the ease of placement, throwing down wires is a lot faster now for sure.

That said, the lack of diagonals and curves is about as limiting as I was expecting it to be, and I don't like how every engine setup is going to be the same box now. It just feels too generic.

The layers will take some getting used to, but are a welcome addition.

What I wasn't expecting was the complete lack of color options for placing wires. Being limited to only 3 colors total feels like another step backward for wiring in general.

I still feel smart wiring would work best as a supplement to old wiring, rather than a replacement. Preserving old functionality would also have the benefit of ensuring map compatibility.

Poojawa:
wires can be colored after setting their layer via multitool. it's just the default colorations otherwise. I made them defines so that if one of the staff wanted the blue layer to be magenta, they can set that define color and *all* cables on Layer 3 are updated, as are the radials, etc. 

I seriously cannot stress enough how impractical the suggestion of maintaining the old behavior alongside the new is. It's not even just a case of setting a var, it's literally creating a second complete system of powernet code and it just doesn't work. 

You're having to double all interactions, all update_icon calls, all propagation checks for something that 99% of the player base won't use and won't notice. All for getting to make squiggly lines at solars or the engine on the rare instance someone feels like spending an hour with the engine completely off in order to safely set up some cables.

FartMaster69420:
> wires can be colored after setting their layer via multitool.

Oh, then I just encountered a bug or got the order wrong there. Glad coloration is still possible.

As for the rest I just have to wonder if what we’re gaining is worth what we’re losing here, do we really need new powernet code?

Zoldorf:
So we just did a test to see how it all works, and the best and most polite way I can put this is that, this does not work, we don't know why or how, but for some reason this has now horribly broken the Field Generators for the tesla engine, they just do not generate a containment field in some tests and in others they do, othertimes emitters do not work.

Cabling is weird and frankly a mess and the new sprites are ugly, I like the layering though.

The new SMES sprite is, in my opinion at least kind of ugly too and very messily and poorly set up on the delight at the very least.

This all also makes me incredibly worried now about ALL the other devices that have not yet been tested to see what even breaks, because having it break essential equipment like the Emitters is not a good sign for other, things either.

I will need to test more of these other new additions and changes too to be able to see what else breaks.

Helmian:
Gave this a look on the test server.  The multiple cable layers is a nice convenience, but the auto-snap and lack of diagonals really is pretty bothersome.  I exclusively play engineer, and adjusting cable management when setting the engine helps keep it at least a little fresh.  While the smart cables will make things go a little quicker and easier, it will also make them a good bit less fun, turning wire work solely into a chore, and making the engine that little bit more of a rote slog.  Aesthetics seem to have been dismissed here a few times, but having your set-up or project look the way you want it to is a decent bit of the joy of it, and the smart cables are, relative to the old ones, rather ugly, along with removing the creativity and occasional puzzle-work of managing power lines.

Overall, while this may make things slightly quicker to do, I believe it will ultimately detract from enjoyment in both personal projects and regular department duties.  I know it certainly will for me.

Poojawa:
I don't believe the test server was actually updated after I pushed the fix, as my own testing allowed me to have a fully operational Tesla going for well over four hours without any issue after I refactored the field generator code. Field generators had this insane snowflake math they were trying to do, and after making the system far more shrimple to compute, they were actually able to maintain their power draw in the surplus that four emitters going would supply. 

da3faada449b7f5c24fba56f4d391cbb


You also do realize the cable sprites are... literally the same sprites, right? the only thing that has changed is the layer offsets. the directionals are all quite literally the same sprites laid over each other...

also the old SMES sprites were incredibly dated. Big white tubes with some emergency ribbing for texture. The current sprites were already in the game, but were squirrelled away on a single POI ship for usage.  I'm uncertain of what you mean by 'messily set up' as well, they only output under themselves, and only input via terminals. You have to switch up how things are laid out in order to have actually sane powernet code. 


Contrary to belief, I'm very glad that stuff is getting fresh eyes on this code. There's only so much I can do on my own and I greatly appreciate your time in digging out thing I overlooked before it goes live.

Zellular:
**[READY] My review of this PR, Poojslop**

sucks

crossexonar:
> **[READY] My review of this PR, Poojslop**
> 
> sucks

Hey, I understand that you're upset about this, but please dont personally attack them over it.

FartMaster69420:
From Xindi on the discord:

> 1. You are unable to build the wire terminals for SMES units, as it eats the wire and produces no terminal on the floor. 
> 2. You can not change the 'wire layer' it seems to accept for its IO, meaning you can't easily edit the SMES.
> 3. APCs seemingly self-create with the ability to use its interface, despite having no cells or circuitry, if the empty APC frame is placed on the same tile as the map starts them at, and the terminals do not disappear when dismantling them. 
> 4. For some reason the engine SMES seems to feeds the MAIN SMES and constantly results in it getting bottlenecked and draining the shield generators to feed everything else instead. Not sure if it was just 'feeding itself' from a mis-mapped wire or not, but both the engine and main SMES were slowly filling up despite only a trickle charge.
> 5. There is seemingly no way to set which of the 'layers' the machines want to use or draw power from if they exist on the same tile (something old wire system could do if you curved the wire instead of placing a knot under it). With no ability to make terminals or choose the layering they need, this breaks a lot of machines/SMES/APCs that require a terminal. 
> 6. Some of the red wires overlap the scrubber pipes and are incredibly difficult to see. Previously not an issue since wires existed offset from them, but very difficult to see with red colourblindness. Same with the blue layer wires and the blue air pipes. 

> 7. The wire layers, in combination with airpipes, obfuscate disposals pipes with all but a 1-wide pixel strip, meaning you have to pixel hunt really hard when interacting/repairing/building disposal pipes or alt-click each tile which takes a few moments to read the tile contents. 
> 8. Despite a SMES outputting 1.30MW, and only charging by 20KW input, the SMES is GAINING CHARGE.  The only 'generator' running at the time is a radiation collector and a single uranium wall.
> 9. After snipping a blue wire from the SMES terminal, I could no longer place it down because an error message says "there's already a cable at that position and layer". Examining the terminal asks for a blue wire. I place a blue wire. It says it is the 'wrong layer' because it's actually asking for a different NUMBERED layer that was also blue, not just any old blue one like the examine text suggests.
> 10. Partially dismantling and rebuilding the SMES spawns additional circuits for the SMES machine.
> 11. It turns out spawning a 'terminal' via buildmode lists 3 distinct terminals, one for each layer, so my money is on the fact the building/construction processes haven't been updated to account for these multiples of terminal types and it just kinda thinks of the older wire code instead of this new one or something, meaning I could never build terminals.
> 12. Machinery that uses the 'heavy power cable' is unable to be connected if not already mapped in, i.e., science can't build PSU banks to charge cells, engineering can't adjust the solar high voltage line, etc Again, due to being unable to build terminals.
> 13. It seems that the station now initialises with the breaker bypasses disabled on the RCON console AND doesn't have the department SMES IO enabled. For a new or inexerienced player they may not know about the RCON stuff, so the station goes dark very quickly. Couple that with the fact the RCON needs power to work, it could be an issue.

image

image

image

FartMaster69420:
Also there's an issue with the default setup on Talon where only the engine only outputs enough power to charge one SMES. The other two will refuse to draw power.

With the old setup SMESes would each draw equal power from the powernet, but it seems now one SMES will draw the full amount leaving nothing for others on the net.

Ichiinu:
So to try and sum up points brought up into one comment for ease of address/evaluation:

What things are added/solved/fixed:

- lighter load on server
- layered wiring allowing separate parallel wiring on one tile (without need for christmas wiring)
- Ota electrical transmitters (do these require the new wiring, or could these be done with old?)

What new/potential problems does this introduce:

- smart wiring with 3+ wiring loops is difficult (could be solved with toggle, but Pooj has said that isn't tenable currently iirc, please correct if I'm wrong)
- diagonal wiring removed
- layering may make circuit readability more difficult (a la atmos piping, but with mixed/solo colors potentially)
- reduces number of independent loop connections you can make on one tile (2 or all 3, vs up to 4 independent pairs) 
- e.g. 

crossedwires
⁠

What's the cost benefit ratio:

- How often are engineers actually encountering situations that necessitate having multiple unconnected wires on one tile? If running parallel, technically doable currently with Christmas light stringing of wires (e.g 
christmaswiring
)
- Not sure on this yet, but important consideration, is what the current engineering crew opinion is on if they actually want the change?
- Aesthetics cost?

Poojawa:
I do not possess the willpower to continue trying to implement this PR as it is. My apologies for the trouble. 

Others can cherry pick assets and functionality as needed to make for a much more atomized and clean implementation that satisfies everyone, as all code is still here via a search.

PR #312, opened 2024-05-18
Cyborg fluffsprite plus change to fluff icon requisites.

foopwo:
A new fluffborg!

Also includes a change discussed with soft to remove checking for an exact match of your borg's name to add custom sprites to the list of options. Since
1. You're prompted to enter name when spawning as borg every time, and it's kind of bad that you get screwed out of your icon if you fat-finger even just a little.
2. Some characters may have naming gimmicks.
3. If you don't want to use your custom icon, like for a different borg character, you can simply not select it.

The ckey check is still there, naturally.


Very-Soft:
you test and make sure this all works correctly?

foopwo:
Yes of course.

PR #316, opened 2024-05-19
Rascal's Pass Engine Satellite

Michab02:
Whoof, big PR. This adds an additional Z-level to Rascal's Pass with a small space station on it, containing everything the engineers need to set up a tesla engine! The satellite can be reached via a shuttle atop the engineering base, which should be relatively easy to find. The station itself has some amenities, but is meant to be extremely small and focused on the engine, in order to encourage people not to linger there and hang out on the station. I considered moving EVA and the landing pads up to the satellite, but turning Rascal's Pass into Tether 2.0 is not the intent of this PR - I just want to add an engine proper to the Pass.

While making this, me and Soft found an issue where Z-level edge transit doesn't work right, and you will sometimes teleport inside the walls down on the pass; this appears to be a larger existing issue not caused by this PR.

image
image


Michab02:
> Honestly, I don’t see the point of this station or the logic. If you want to use it to power Rascal Pass, then it’s not even logical, as if you have a wire hanging from the earth to space, which sounds crazy. And since there is no pipe from earth to space, it looks even more strange and out of place. It is better to transfer the developments to a new space station which was planned.

It's using fictional wireless bluespace transmission. Don't worry about it.

Michab02:
The 'lore' can be worked on in a second draft, once the satellite has proven to be successful with the crew. I'm not too worried about the fluff right now.

PR #329, opened 2024-05-20
adjusts enginesat for singulo testing

Michab02:
after Xindii tested it it turns out the singulo's EMP is powerful enough to shut down the entire station, so this alters the APC to be further back in order for testing to happen

there is an issue where turfs on the enginesat appear to default to sand tiles, which may?? cause the singulo to eat the floor and grow infinitely?


Michab02:
this was initially just a patch to remove the singularity gen parts outright, but xindii requested I leave them there for testing purposes

PR #335, opened 2024-05-21
Removes ability for research cyborgs to self-upgrade.

foopwo:
Did you know that research cyborgs can trivially shove a reset board into themselves and become any other module on the fly? Let's nip that in the bud.
Self-repairing is disallowed, after all, self-upgrading is all too similar.

Also disallows self-unlocking because you know they'll just resort to dragging around a crate of upgrades and telling the first assistant they find to bonk it with them.

Does **not** prevent research cyborgs from unlocking and upgrading others.

### DIFF
--- a/code/modules/mob/living/silicon/robot/robot.dm
+++ b/code/modules/mob/living/silicon/robot/robot.dm
@@ -707,6 +707,8 @@
 				update_icon()
 			else
 				to_chat(user, "<span class='filter_notice'><font color='red'>Access denied.</font></span>")
+				if(user == src) //RS edit: No self-unlocking.
+					to_chat(user, "<span class='filter_notice'>You are not allowed to manipulate your own lock.</span>")
 
 	else if(istype(W, /obj/item/borg/upgrade/))
 		var/obj/item/borg/upgrade/U = W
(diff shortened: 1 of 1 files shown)


FartMaster69420:
Is this something that's a problem?

code/modules/mob/living/silicon/robot/robot.dm
```diff
@@ -846,7 +850,10 @@
 	else if(istype(M, /mob/living/silicon/robot))
 		var/mob/living/silicon/robot/R = M
 		if(check_access(R.get_active_hand()) || istype(R.get_active_hand(), /obj/item/weapon/card/robot))
-			return 1
+			if(R == src) //RS edit: No self-unlocking.
+				return 0
+			else
+				return 1
```
Poojawa, line 856:
```suggestion
			if(R == src) //RS edit: No self-unlocking.
				return FALSE
			else
				return TRUE

```

It'd be good to be in the habit of cleaning up old files as we go along to use actual TRUE/FALSE to remove more magic numbers.

code/modules/mob/living/silicon/robot/robot.dm
```diff
@@ -707,13 +707,17 @@
 				update_icon()
 			else
 				to_chat(user, "<span class='filter_notice'><font color='red'>Access denied.</font></span>")
+				if(user == src) //RS edit: No self-unlocking.
+					to_chat(user, "<span class='filter_notice'>You are not allowed to manipulate your own lock.</span>")
```
Poojawa, line 711: this feels like the check is too late, did you confirm that this does actually toggle the self locking? because if they're allowed to swap people's locks (which includes themselves) then the else part of the statement won't apply.

foopwo: This provides additional feedback after check_access fails. It's mostly here because it's where all the rest of the messages are.

lm40:
You lose most upgrades if you reset module, and I don't really see this as a problem ATM.

foopwo:
You keep vtec and scrambled module, actually.

PR #345, opened 2024-05-26
Bone related

Very-Soft:
An idea for a problem - Requires discussion with staff

When a bone is generated via digestion, it registers the predator to a variable, and begins listening to the signal for the predator's movement proc, and destroy proc.

When the predator moves, if the bone is left on a turf, and if the predator is more than 7 turfs away from the bone on x or y, and also in less than 25% light, then the bone will delete itself. The bone will not delete itself unless it is on a turf.

Additionally, makes bones delete themselves when you are deleted. This way if you cryo, all of the bones registered to you, no matter where they are, will be removed. There isn't really a difference between cryoing and other methods of being deleted, so, things like digestion and gibbing will also result in the bones being deleted, but there's not a whole lot I can do about that.

If a bone is picked up by someone, the bone will become bonded to the new person who picked it up, so it will then only delete itself if the new person moves out of range of it, or is deleted as described above.

Right now the bones just instantly disappear. Maybe we can make them display a message and fade out or turn into ash instead if just instantly disappearing.

The idea behind this is to have mechanics better represent the intended use case for these mechanics, and hopefully resolve a long standing issue with the digestion remains mechanic. Some people like it, some people can't stand it. Rather than making convoluted and difficult to define and follow rules about how the mechanic can be used, I am seeking a mechanical solution to the problem.

The intention for these bones is, that you can have them around you, you can hide them away somewhere, you can give them to people. You should not leave them out in the open somewhere for others to find.

### DIFF
--- a/code/modules/vore/eating/leave_remains_vr.dm
+++ b/code/modules/vore/eating/leave_remains_vr.dm
@@ -100,6 +100,7 @@
 	w_class = ITEMSIZE_SMALL
 	var/pred_ckey
 	var/pred_name
+	var/mob/living/pred
 
 /obj/item/weapon/digestion_remains/synth
 	name = "ruined component"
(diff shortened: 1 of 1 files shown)


Michab02:
whilst I am not staff, I do like this change in theory. It allows people to freely play with mechanics while mitigating (if not removing) the existing pref issues with bones.

CockatriceXL:
I'm worried that the 7 turf limit would prevent some engaging situations, such as a predator making a lair somewhere in maint or a cave and amassing a collection of bones. They wouldn't be able to go out to hunt since the bones would vanish when they get too far away.

Issue #356, opened 2024-05-30
[GENERAL] Adranol's been broken for 7 years


ExabyteDiscord:
#### Brief description of the bug
Adranol causes jitters

#### What you expected to happen
Adranol stops the jitters

#### What actually happened
Adranol increases jitters exponentially

#### Steps to reproduce
- (Step 1) Drink covfefe
- (Step 2) administer adranol
- (Step 3) Become perpetual motion machine as your jitters never cease

#### Code Revision
- all of them

#### Anything else you may wish to add:
https://github.com/TS-Rogue-Star/Rogue-Star/blob/e6632c371fadabd20ccb5a767bcacdcacd3fc5d7/code/modules/reagents/reagents/medicine_vr.dm#L18
This should be M.make_jittery(min(-25*removed,0)) because the current code actually just takes the amount of jitters they have, subtracts it by 5, and then adds it to them until they hit the cap of 1000.

If you're wondering why I haven't just fixed this myself it's because I cannot remember how to get gitkraken or github to work so we work with what we got. Also submitted to vorestation's github [here.](https://github.com/VOREStation/VOREStation/issues/16014)

PR #363, opened 2024-06-04
Port Adranol Fix

Durandaal:
Ports [VOREStation PR16015](https://github.com/VOREStation/VOREStation/pull/16015). 

Adranol now functions as intended; clearing a user's jitters in half the time.

### DIFF
--- a/code/modules/reagents/reagents/medicine_vr.dm
+++ b/code/modules/reagents/reagents/medicine_vr.dm
@@ -15,7 +15,7 @@
 	if(M.eye_blurry)
 		M.eye_blurry = max(M.eye_blurry - 25*removed, 0)
 	if(M.jitteriness)
-		M.make_jittery(max(M.jitteriness - 25*removed,0))
+		M.make_jittery(min(-25*removed,0)) //RSEdit: Ports VOREStation PR16015, makes Adranol function properly
 
 /datum/reagent/numbing_enzyme
 	name = "Numbing Enzyme"


foopwo:
Forgot to tag a 'fixes #356' in this

code/modules/reagents/reagents/medicine_vr.dm
```diff
@@ -15,7 +15,7 @@
 	if(M.eye_blurry)
 		M.eye_blurry = max(M.eye_blurry - 25*removed, 0)
 	if(M.jitteriness)
-		M.make_jittery(max(M.jitteriness - 25*removed,0))
+		M.make_jittery(min(-25*removed,0)) //RSEdit: Ports VOREStation PR16015, makes Adranol function properly
```
Poojawa, line 18: If you're porting a fix from upstream, you don't need to add an 'RSedit' because...it's an upstream fix???

Durandaal: As a matter of course, any code that isn't written by me gets an attribution tag. Maybe it isn't needed in this instance, but eh.

Poojawa: simply making sure you credit the PR (which you did) is fine enough for accreditation, especially for low level/ simple fixes like this. Git's blame will show that you were the one to have changed this line and people will be able to backtrace it through the this PR number!

PR #368, opened 2024-06-11
Makes unsimulated planetary turfs much better at changing temperature.

foopwo:
This works by giving unsimulated planetary turfs an effective boost to their heat capacity to the same degree that unsimulated turfs have for gas mixture. Specifically done for planetary turfs as to not change current behavior with space turfs.

I do not know for certain how much extra processing this adds, through these checks are only done once per tick for every zone interacting with unsimulated turfs, and only when there's a difference worth processing, so it shouldn't be much.

Rudimentary tests were done with various things such as active supermatters, venting toxins burn chambers to the air, popping canisters with gas mixtures varedited to ridiculous levels, but I feel this warrants a 'hey everyone go out of your way to mess with the atmosphere, go wild' test round on rascal's pass before merging.

### DIFF
--- a/code/ZAS/ConnectionGroup.dm
+++ b/code/ZAS/ConnectionGroup.dm
@@ -224,9 +224,9 @@ Class Procs:
 		erase()
 		return
 
-	var/is_planet = 0 // RS edit
+	var/is_planet = FALSE // RS edit
 	if(istype(B,/turf/unsimulated/wall/planetary))
-		is_planet = 1
+		is_planet = TRUE
 	var/equiv = A.air.share_space(air, planetary = is_planet)
 
 	var/differential = A.air.return_pressure() - air.return_pressure()
--- a/code/ZAS/ConnectionGroup.dm
+++ b/code/ZAS/ConnectionGroup.dm
@@ -224,7 +224,10 @@ Class Procs:
 		erase()
 		return
 
-	var/equiv = A.air.share_space(air)
+	var/is_planet = 0 // RS edit
+	if(istype(B,/turf/unsimulated/wall/planetary))
+		is_planet = 1
+	var/equiv = A.air.share_space(air, planetary = is_planet)
 
 	var/differential = A.air.return_pressure() - air.return_pressure()
 	if(abs(differential) >= vsc.airflow_lightest_pressure)
--- a/code/modules/xgm/xgm_gas_mixture.dm
+++ b/code/modules/xgm/xgm_gas_mixture.dm
@@ -394,7 +394,7 @@
 
 
 //Shares gas with another gas_mixture based on the amount of connecting tiles and a fixed lookup table.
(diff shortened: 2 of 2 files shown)


code/ZAS/ConnectionGroup.dm
```diff
@@ -224,7 +224,10 @@ Class Procs:
 		erase()
 		return
 
-	var/equiv = A.air.share_space(air)
+	var/is_planet = 0 // RS edit
+	if(istype(B,/turf/unsimulated/wall/planetary))
+		is_planet = 1
```
Poojawa, line 229:
```suggestion
	var/is_planet = FALSE // RS edit
	if(istype(B,/turf/unsimulated/wall/planetary))
		is_planet = TRUE
```

code/modules/xgm/xgm_gas_mixture.dm
```diff
@@ -394,7 +394,7 @@
 
 
 //Shares gas with another gas_mixture based on the amount of connecting tiles and a fixed lookup table.
-/datum/gas_mixture/proc/share_ratio(datum/gas_mixture/other, connecting_tiles, share_size = null, one_way = 0)
+/datum/gas_mixture/proc/share_ratio(datum/gas_mixture/other, connecting_tiles, share_size = null, one_way = 0, planetary = 0)
```
Poojawa, line 397:
```suggestion
/datum/gas_mixture/proc/share_ratio(datum/gas_mixture/other, connecting_tiles, share_size = null, one_way = FALSE, planetary = FALSE)

```

code/modules/xgm/xgm_gas_mixture.dm
```diff
@@ -441,8 +443,8 @@
 
 
 //A wrapper around share_ratio for spacing gas at the same rate as if it were going into a large airless room.
-/datum/gas_mixture/proc/share_space(datum/gas_mixture/unsim_air)
-	return share_ratio(unsim_air, unsim_air.group_multiplier, max(1, max(group_multiplier + 3, 1) + unsim_air.group_multiplier), one_way = 1)
+/datum/gas_mixture/proc/share_space(datum/gas_mixture/unsim_air, planetary)
+	return share_ratio(unsim_air, unsim_air.group_multiplier, max(1, max(group_multiplier + 3, 1) + unsim_air.group_multiplier), one_way = 1, planetary = planetary)
```
Poojawa, line 447:
```suggestion
	return share_ratio(unsim_air, unsim_air.group_multiplier, max(1, max(group_multiplier + 3, 1) + unsim_air.group_multiplier), one_way = TRUE, planetary = planetary)

```

Poojawa (commented):
Continuing purges of magic numbers for TRUE/FALSE wherever they crop up.

Unless these have more than one on/off state anyway.

PR #371, opened 2024-06-14
Revises Hypospray mkii performance

Poojawa:
After obtaining feedback over a lengthy period of time, this PR is set to ensure that syringes have actual competition in medicine application.  Delay was greatly reduced to use the hyposprays as the application time way was longer than intended, The kits can now also hold 14 vials at once, on par with a regular medkit holding 7, 60u bottles.

It also greatly improves availability of hypospray mkiis and kit boxes by adding a lot of empty ones to the nano med, hypovials can be produced from the autolathe as well.

Adds the CMO's version of the mkii into their locker, as well as bundles up the mki version into a box with some extra vials for those who prefer the older over the new one. The CMO's mkii holds 60u 'large' vials and is even faster than regular mkiis. As well as being able to apply larger doses all at once, comparable to pills. 

Also removes the extra pair of white shoes because clothing vendors exist.


FartMaster69420:
Balancing the hyposprays around be available at round start is doomed to fail in one of two ways.
A) Hypos are balanced too powerful and negate the use of syringes outside of flavor/edge cases
or
B) The hypos are nerfed to the point there is little reason to use one over a syringe

I feel these should be a higher tier item built by science, if it's locked behind a rare or harder to unlock tech outside of the 'meta' rush. EG: Require deconstruction of a phasic health scanner, or anomaly/precursor tech then we don't have to worry about the hypos becoming too widespread for every round and can be allowed to be a straight upgrade.

Further, I'm not sure we need a new type of beaker for the hypos, regular vials are already only used for the CMO Hypos, the new vials just add more item bloat.

Poojawa:
I disagree on locking them behind science. They're as good as never-to-be-used at that point. They cannot be built in science or ordered from cargo, so there's a finite amount and that's okay. Putting them behind such a massive tech wall is essentially just telling me to just remove them entirely from the code and I don't really want to do that. 

They've already been available at roundstart for a quite a while now, and the feedback I received was the delay being too excessive that it was still better to simply harm stab someone three times with a syringe in the time it took one mkii injection to occur. The other feedback was the advantage that just carrying 60u bottles meant you had more on hand, and you could just carry a bunch of medkits filled with bottles, so now hypokits are up to par with the same capacity. 

Syringes still have their advantages, and I'm not going to fiddle with them because I don't want to. 

The different subtype was specifically because vials are used by virology machinery, and I intended to make hypovials only for the mkii to use.

Durandaal:
A smol nitpick: the CMO's Mk.II hypospray kit shouldn't have "antitoxin", it should be renamed to "Dylovene" to bring it in line with the other chems in the kit. Otherwise, I like the changes at first brush.

Poojawa:
Good catch, swap'd 'em over. Also updated the pathing for clarity

Poojawa:
Reopening as this is still a valid PR and I'd just like to have the 'This is still bothering me it isn't fixed' impulse out.

code/modules/economy/vending_machines.dm
```diff
@@ -408,11 +408,12 @@
 					/obj/item/weapon/reagent_containers/glass/bottle/toxin = 4,
 					/obj/item/weapon/reagent_containers/syringe/antiviral = 4,
 					/obj/item/weapon/reagent_containers/syringe = 12,
-					/obj/item/weapon/reagent_containers/glass/bottle/hypovial/small = 12,
-					/obj/item/weapon/hypospray_mkii = 4,
+					/obj/item/weapon/reagent_containers/glass/bottle/hypovial/small = 40,
+					/obj/item/weapon/hypospray_mkii = 10,
+					/obj/item/weapon/storage/firstaid/hypokit/empty = 10,
```
Poojawa, line 413: This can be changed to just 10 kits, but I figured the modularity aspect would have been preferred.

PR #379, opened 2024-06-17
Promethean Internal Bleed Fix

FartMaster69420:
Prometheans getting internal bleeds doesn't make a lot of sense, so this PR seeks to resolve this issue.

Rather than make prommies immune, I added the ability for their limb regeneration ability to repair internal bleeds so they still need to worry about the injury, but it shouldn't snowball into instant death as easily.

Prometheans will not fix external bleeds this way, and will still need standard medical intervention for regular wounds and blood levels are unaffected by regeneration so transfusions will also be required.

### DIFF
--- a/code/modules/mob/living/carbon/human/human_powers.dm
+++ b/code/modules/mob/living/carbon/human/human_powers.dm
@@ -346,6 +346,13 @@
 
 		handle_organs() // Update everything
 
+		for(var/obj/item/organ/external/O in src.bad_external_organs)
+			for(var/datum/wound/W in O.wounds)
+				if(W.internal)
+					W.damage = 0
+					O.wounds -= W
+					to_chat(src, "<span class='notice'>You feel a tightening sensation as [O.name] clots.</span>")
+
 		update_icons_body()
 		active_regen = FALSE
 	else
--- a/code/modules/mob/living/carbon/human/human_powers.dm
+++ b/code/modules/mob/living/carbon/human/human_powers.dm
@@ -351,7 +351,7 @@
 				if(W.internal)
 					W.damage = 0
 					O.wounds -= W
-					to_chat(src, "<span class='notice'>You feel a tightening sensation as [O.name] clots.</span>")
+					to_chat(src, "<span class='notice'>You feel a tightening sensation as your [O.name] clots internal damage.</span>")
 
 		update_icons_body()
 		active_regen = FALSE


crossexonar:
Hey, I appreciate you looking out for prommies but consider, this makes them even less reliant on actual medical treatment, so not sure how to feel about this one

FartMaster69420:
While that’s true, the original intent is for prometheans to be resistant to brute and weak to burns. At the moment they can tank enough brute that they’ll pretty much jump to the highest level of IBs which leads to deaths that are too rapid for medical to respond too.

I’ve literally had a situation where I bled out and died on the operating table as they were trying to fix the bleed.

I’m open for alternative balance, but the IBs as they are hit too hard.

BriarGarden:
Since it was brought up, I'm actually pretty against this idea without an alternative that'll actually kill the promethean. In my experience playing them, they don't care about medical under all but the most extreme conditions (ie; active firefight with syndi mobs).

As it currently stands, having a medkit, something to eat (possibly syndicakes even), and playing a bit carefully--you're functionally immortal. No broken bones, no infection, no limb loss that stops you from continuing, literally absolute no need for medical's intervention at all. Which is already ridiculously potent, removing the one thing that actually does kill them for being too aggressive is not a balance change I support at this time. 

So, I'm going to put my stance as firmly **against** a merger unless there's something to replace it.

FartMaster69420:
Prometheans are already punished severely for being too aggressive by their weakness to lasers. A combat drone or laser hivebot can put a prommie into crit with 2-3 shots with the potential of stun locking in 1 shot if they hit a leg.

This weakness to burns is meant to be balanced by a resistance to brute, but the current implementation of internal bleeding effectively gives prometheans weakness to brute as well as their weakness to burns.

As far as medical interactions, for the majority of my Promethean internal bleed experience it starts and ends with injecting phoron into the slime core.

FartMaster69420:
I should note, this is not a removal of internal bleeds. It just makes the treatment for them consistent with the rest of slime ailments, which usually amounts to feed the slime so they can regenerate.

BriarGarden:
Their strength verses brute and weakness to burn has nothing to do with removing the only thing that actually kills 'em and/or gets medical involved. Anyone can pick up brute resist, and even pair it with burn weakness if they so choose, but they also have to deal with broken bones, removal of limbs, internal bleeding, and the high risk of infection if they just cast it and keep going. In other words, without medical intervention, death. Usually fairly quickly too.

Even Synths take enough damage to have internal breaks, and limb removal they can't quite overcome and need to be thrown down on a surgery table. Prometheans should not be excluded from an entire aspect of the station and game itself, when they already _mostly_ do. I'll state it again, without something to replace this and have them in medical to some capacity I'm staying firmly in the 'No' camp.

FartMaster69420:
I don’t know what to tell you other than your take that Prometheans are unkillable outside of internal bleeding does not match my experience in playing Prometheans.

You also seem to be forgetting Proteans, a species that requires no aid outside of being placed into the recently added revival machine on the rare cases they die.

And for that matter, Xenochimeras are quite literally unkillable with their revival system.

crossexonar:
Uno dies a lot from what I've run alongside with him as sari, I also die a lot, in either nanogoop or real goop, the IB just sneaks up on you out of nowhere and theres not a damn thing you can do about it unless you powergame prepare for it in advance Or get lucky with a bicar overdose.

FartMaster69420:
Closing this as from chatter in Dev chat it may be best to rebalance internal bleeding across the board, and I'd also like to do something funny with prometheans when they run out of blood.

Both of which go outside the scope of this PR.

PR #383, opened 2024-06-19
FasterNomNotifications!

Lilly-Lira:
The powers is within u

### DIFF
--- a/code/modules/mob/dead/observer/observer_vr.dm
+++ b/code/modules/mob/dead/observer/observer_vr.dm
@@ -69,7 +69,7 @@
 	if(db)
 		var/datum/transhuman/mind_record/record = db.backed_up[src.mind.name]
 		if(!(record.dead_state == MR_DEAD))
-			if((world.time - timeofdeath ) > 5 MINUTES)	//Allows notify transcore to be used if you have an entry but for some reason weren't marked as dead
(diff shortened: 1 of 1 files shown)


Arokha:
When you discussed this you said "For vore deaths", this will apply to all deaths or even just manual ghosting, so someone could manually ghost and immediately ping with this change.

To explain what's happening, the reason you can't immediately ping is because there's two time-sensitive events happening. One is the transcore sweeping over the backup implants every *n* minutes, and checking when the implant last sent a ping. If it's greater than the overdue time, it marks them as dead (`transcore_vr.dm:120`). At that point you can immediately use the notify verb.

The other time sensitive thing is that if you died more than 5 minutes ago, this proc you're editing allows you to ping if for some reason the transcore didn't mark you as dead. That should never really happen, the transcore sweeps happen more often than every 5 minutes I believe, but this is meant more as a catch-all thing for situations where the transcore sweeps stop happening (like the subsystem breaks or whatever).

What might be better for what you want, assuming you want it "for vore deaths", is to just manually set the record to MR_DEAD on gurgle death. That way the verb is immediately usable by them, but not by manual ghosts, people who died in space, etc. They have to wait a few minutes.

Lilly-Lira:
Thanks!  This seems like a better way to do it.  I will update :3

PR #393, opened 2024-06-23
Osteodaxon Rebalance

FartMaster69420:
### **Lore Dump:** 

Due to over use of Osteodaxon by NT Personnel a sharp decrease in drug effectiveness has been detected! We suspect this is caused by growing crew tolerance to the drug, and to combat this we recommend increased dosages! Do be aware, there may be some unwanted side effects, so it is recommended crew are anesthetized while undergoing treatment.

### Standard PR Info:

**The Problem:** 

Osteodaxon is an easy chem to obtain, and it is incredibly efficient, with .1 units being enough to treat every broken bone in someone's body.

Recently, medical has found several speed running strats to ensure it can be obtained without even the minimal effort of engaging with space carp.

This is causing issues where surgeons are being pushed to the wayside from the one injury they can reliably encounter as the pill fixes it faster and with no drawbacks.

**The Solution:**

Osteodaxon has been nerfed across the board in this PR, specifically it was adjusted to cause HalLoss while processing, and a dosage of 4 units is needed to fix broken bones if injected. Pain given will be greater than even ParaTram can counter, requiring the patient be put under while metabolizing the chem or to be administered oxycodone. Even then, the halloss is likely to put them into crit for a short time.

When applied as a pill Osteo will require 8 units to process before bones are fixed, but as a trade off it causes less halloss and the pain levels are manageable with ParaTram.

Further, 15 units of halloss are applied at the final repair and pain levels for the bone repair message were buffed.

### TL;DR

- Osteo no longer fixes bones at any dosage.
- As an injection 4 Units are required, this allows for easy dosage from a bottle to a syringe.
- Pain and Halloss are applied, enough to put the patient into paincrit without medical attention.
- As a pill, 8 Units are required, however less pain is given per tick and the pain is low enough to be treated with ParaTram.
- When fixed, 15 Halloss will be applied.
- 4 Units of HalLoss are applied as it metabolizes in the bloodstream, building up paincrit quickly
- 2 Units of HalLoss are applied as it metabolizes in the stomach, making it the lower pain option with a longer time to finish and higher chem cost.

### DIFF
--- a/code/modules/reagents/reagents/medicine.dm
+++ b/code/modules/reagents/reagents/medicine.dm
@@ -711,11 +711,11 @@
 	overdose_mod = 1.5
 	scannable = 1
 
-/datum/reagent/osteodaxon/proc/FixBones(var/mob/living/carbon/human/H)
+/datum/reagent/osteodaxon/proc/FixBones(var/mob/living/carbon/human/H) // Pushed off to it's own proc since we'll need to use it both for ingest and blood.
 	for(var/obj/item/organ/external/O in H.bad_external_organs)
 		if(O.status & ORGAN_BROKEN)
 			O.mend_fracture()		//Only works if the bone won't rebreak, as usual
-			H.custom_pain("You feel a terrible agony tear through your bones!",150)
+			H.custom_pain("You feel a terrible agony tear through your bones!",80)
 			H.adjustHalLoss(15) // Fixing bones hurts like fuck
 			H.AdjustWeakened(1)		//Bones being regrown will knock you over
 
@@ -730,16 +730,16 @@
 		if(rand(1,3) == 3)	// We don't want to spam pain messages
 			switch(rand(1,5))
 				if(1)
-					H.custom_pain("It feels like your bones are on fire!", 120)
+					H.custom_pain("It feels like your bones are on fire!", 120) // While more efficient, we get a LOT more pain.
 				if(2)
-					H.custom_pain("You feel your bones shifting under your skin!", 120)
+					H.custom_pain("You feel your bones shifting under your skin!", 120) // This pain is enough that Oxycodone or anesthetic will be required.
 				if(3)
 					H.custom_pain("Your bones ripple and contort agonizingly!", 120)
 				if(4)
 					H.custom_pain("Every bone in your body is screaming for help!", 120)
 				if(5)
 					H.custom_pain("Your bones feel as if they're fighting against your every movement!", 120)
-		if(dose >= 4) // We need four units for repairs to happen.
+		if(dose >= 4) // In blood osteo is more efficient
 			M.heal_organ_damage(3 * removed, 0)	//Gives the bones a chance to set properly even without other meds
 			FixBones(H)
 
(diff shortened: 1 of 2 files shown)


FartMaster69420:
I wouldn't consider the numbers for dosage and halloss final, they may need tweaking depending on how it feels in live play.

FartMaster69420:
Note from testing Carp give approximately 30 Units of Osteo per standard carp, assuming butchered messily.

Combat Kits give 15 units per pill, in 3 pills. So from a single combat kit medical would have 45 Units of Osteo.

lm40:
Okay, so, I was already thinking about what changes to make to osteodaxon myself. I don't really think this is the approach that I am going to choose, but we will see. At any rate, I am not seriously considering to make a change like this in the coming weeks, let alone the coming days. I was led to believe that you made this PR fully accepting the possibility of it being wholesale rejected, and so, I am fine to leave it open until I am ready to visit the topic of osteodaxon balance. There are some other more proximate ideas to be tried first.

keane0tts:
Personally, I like this change. Would have been nice to play around  with.

FartMaster69420:
You can download the branch and run it on a test server if you wanna play around with it, Keane.

Even if this never gets merged feedback on this change will help shape any future rebalances done.

tachigami-arch:
not a fan, to be honest. This seems like a gross overcorrection. Pill form vs injection form differences just makes the stuff far more complicated than it needs to be, and removes the option to use pills for medics because none of us are going to want to waste literally half our supply just for less pain, and not even enough less pain not to be roleplay-significant, while _also_ taking longer. Osteo only ever really comes in two situations: we have a little bit, or we have functionally all of it anyone could ever need. This change would eliminate the former completely, and make the latter more of a pain and prevent anyone from having rezadone more than anything else. I don't think anyone would use osteo outside of combat kits in this situation.

Ideally I'd suggest as a different way to do it, keep osteo as it is. Make these changes instead: It metabolizes instantly after a small period, and when it finishes metabolizing it deals pain inversely proportional to how much osteo was metabolized. 1 unit gives a message on how much it hurts, equivalent to what it currently does, ~0.2 or lower puts them into crit for a minute and slows for a minute after that, ~0.4 crits them for 45 seconds or so, ~0.6 15 seconds, ~0.8 a few seconds, and going up to 2u only causes minor pain and discomfort. This would allow for osteo to be used at lower values and keep the ability to have any usable amount in the "we have very little" circumstances, while also bringing up a more serious question as to medical options when it comes to what patients might want to ask for knowing osteo would have a more significant pain effect at lower strengths. It would also encourage medics to still desire more osteodaxon and have reasons to trade with talon and send people on carp/shark hunts even when a small amount of osteo is already in their possession.

idk, just seems better than effectively removing the stuff outside of large quantities, it can be very useful even in circumstances where people are already doing plenty of surgery, like repairing corpses for resuscitation.

FartMaster69420:
I agree that the current unit requirements here may be too high, this PR requires proper playtesting to iron things out, but leaving the option of 0.1 pills, even with pain as a downside, would not solve the core issue.

Under the current balance, 5 Units of Osteo can quite literally be turned into 50 Pills of equal effectiveness. And these are pills that can be made just after round start without interacting with any other department and with little to no risk.

lm40:
I didn't really want this then and since all the medical re-balancing frenzy seems to have died down I think it is even less necessary now

PR #399, opened 2024-06-24
Portal stuff (untested)

Arokha:
- Adds a var, `notify_ckey_once`, if set to a ckey will give that user a tgui alert the next time (and only the next time) something goes through the portal, with the x,y,z of the portal used, and the name of what was portal'd
- Adds a var, `open`, which allows one to toggle whether the portal is portaling right now or not
- Adds a var, `close_after_uses`, after portaling this many things/people, the portal will close.


Very-Soft (changes requested):
Conflicts

PR #404, opened 2024-06-27
Fix river leech

Arokha:
The implant removal surgery makes special account of borers, but not river leeches. This fixes the 'still connected to host over infinite distance'.

Additionally tweaks things about leeches: I think they were intended to be player-controlled because there are some things that apply to player-controlled ones (cordradaxon making them docile only applies to players and not AI based on how the code was written).

Now, if you inject a host with cordradaxon, wait one life tick (so a few seconds) and then remove it, it will be removed while docile, which prevents any special ability use by player or AI control. So, should have a much easier time murdering it. They will remain docile for 60 seconds if removed while docile.
(Untested)


SigholtStarsong:
RIP to my 5th-dimensional leech homies. Now more infinite-range succ.

PR #407, opened 2024-06-28
Surgery Buff!

FartMaster69420:
### Intro

So, surgery has two problems right now.

1) It's incredibly slow to preform
2) In field situations you'll often fail steps, and make the situation worse at the mercy of RNG.

So in this PR I aim to fix both issues.

What this does is remove the RNG failure from surgery and replace it with a speed multiplier for surfaces.

Now, instead of failing a step you must work longer in field conditions, and in ideal conditions you get a significant speed boost to surgery.

### Numbers
Operating Tables get a 3x boost to surgery speed.
Tables preform surgery at half default speed.
Roller beds preform surgery at 75% default speed.
Advanced Roller beds preform surgery at 1.2x default speed.

### Technical
The surgery_odds variable has been replaced with a surgery_mult variable, which is then used to divide the base surgery speed. So a higher number in surgery_mult is a faster surgery, 1 is default speed, and decimals below 1 are slower speeds.

Currently in code, the only surfaces that can preform surgery are tables, roller beds, and operating tables. This could be expanded simply by adding a surgery_mult value above zero to any surface desired to work as such.


crossexonar:
Does this affect robotics surgery too?

Surgery being the main thing I actually enjoy to do in medical, faster is cool.

Advanced surgery tools work with this?

FartMaster69420:
Yes, this effects all surgeries including robotics.

Tools speeds also stack.

Skysteak:
I am not sure how I feel about the field surgery changes, but I **fully** support the conventional surgery changes. If you need to make it _slightly_ slower (5-6 seconds for base tools),  for the advanced ones to shine? That's fine too.

FartMaster69420:
Yeah, splitting this into a different PR may be preferable. I'll wait to see what maintainers think, but there's a good chance I'll do that.

My intent here is to remove RNG mechanics from field surgery and to replace them with slower operating speed, since running surgery in the field has a bit of a slot machine feel to it that just isn't that fun.

Ideally, I'd like to make field surgery worse without relying on RNG.

crossexonar:
actually, please yes increase surgery speed, even if only for corpsemend, make it actually compete with the ml3m gun

crossexonar:
on further thoughts, roller beds should NOT make surgery move faster, if anything it should be slower than a table which is a sturdy surface.

FartMaster69420:
Okay, pushed an edit that nerfs the Roller bed to 0.5 times speed, and buffs the tables to 0.75 times speed.

The Advanced Roller Bed was left at 0.75 since I'm assuming it has advanced stabilizing tech in it, and if you have an advanced roller bed you can probably also build a table.

Also fixed vorestation.dme being included in the edits in this PR, it shouldn't have been there to begin with.

FartMaster69420:
> Change regular table to .25 times speed and start off regular surgery at 2 times speed and I will accept this. I get the sentiment that table should be more stable than rollerbed but tables are everywhere and just from a game perspective I rather give better numbers to the actual medical equipment. These values will not be set in stone and as play develops I will probably tweak them.

Should roller beds remain unchanged? This change makes the advanced Roller Bed a bit close to the surgery table for speed. (2x mult vs 1.2x mult)

FartMaster69420:
> I thought advanced roller beds were at .75, the full list of what I want is table at .25, regular roller bed at .5, advanced roller bed at .75, and proper surgery table at 2.

Done.

I've also included a file that defines the multipliers so they can be easily changed without touching each surface individually.

Defines are in code/__defines/medical_RS.dm

lm40 (approved):

PR #408, opened 2024-06-29
Confetti Cannon Port

Durandaal:
Ports the confetti cannon from [VOREStation PR15493](https://github.com/VOREStation/VOREStation/pull/15493) and [CHOMPStation PR530](https://github.com/CHOMPStation2/CHOMPStation2/pull/530). This can be researched by science at combat level 2 and material level 2. It can be loaded with paper for confetti, banana peels, or banana creme pies for slipping or splat shenanigans. 

Also contains two admin spawn items, the turbo version and the robot version. The turbo version has a faster rate of fire, and the robot version does not require ammunition and can select a given ammunition type at-will.


Skysteak:
It'd be an utter _slip up_ not to have this here! Some may find pie throwing to be objectionable and crass, but it is in fact the realm of the _upper crust_! All in all, may a colourful device, and may success rain down upon you!

PR #413, opened 2024-07-02
Fix gateway & EVA access

Michab02:
fixes the explo gear EVA access to the proper way (I messed it up last time), and gives the _vr version of research director gateway access- previously they were accidentally excluded from accessing the gateway on vorecode only.


vorestation.dme
```diff
@@ -4300,7 +4300,6 @@
 #include "maps\expedition_vr\beach\submaps\mountains.dm"
 #include "maps\expedition_vr\beach\submaps\mountains_areas.dm"
 #include "maps\gateway_archive_vr\blackmarketpackers.dm"
-#include "maps\groundbase\groundbase.dm"
```
Poojawa, line 4303: regression here, by the by

Michab02: HECK

PR #415, opened 2024-07-02
Restore body markings on protean refactor

Arokha:
Should restore protean markings when they rebuild parts of themselves. Easiest on-mob source for it is in dna, which should have a copy. Alternatively we could get it from character setup, but if you load a different slot, then maybe shenanigans would happen.

Also fixes body color keeping


crossexonar:
Testing COMPLETE, I helpied!

lm40 (approved):

lm40:
Looks like it works to me.

PR #416, opened 2024-07-02
Fixes markings for Prommies and Diona

FartMaster69420:
Reapplies markings after regenerating limbs for Prometheans and Diona.

Original code from @Arokha


lm40 (approved):

lm40:
Looks like it works to me.

PR #420, opened 2024-07-03
Fixes collar descriptions

Arokha:
Probably! Untested.

### DIFF
--- a/code/modules/client/preference_setup/loadout/gear_tweaks_rs.dm
+++ b/code/modules/client/preference_setup/loadout/gear_tweaks_rs.dm
@@ -1,0 +1,12 @@
+/*
+Custom Description (for collars)
+*/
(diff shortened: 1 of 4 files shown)


lm40 (dismissed):

lm40:
Probably.

Arokha:
Probably! Lex/Rykka said she'd help test later so I'd just leave this alone until then.

lm40:
So actually it doesn't work on spawn only on rename, therefore I will RETRACT my approval.

Arokha:
Got un-lazy and set up a test environment. Fixed issue with spawning, tested spawning, renaming, etc.
I also removed the loadout 'name' customization, as it has never worked (it's created procedurally). It seems like a bad idea to have anyway.

lm40 (approved):

PR #429, opened 2024-07-09
Protean Species and Abilities Overhaul [Feedback/Testing]

crossexonar:
Hello dear readers, I have played as a protean for a long time (Back since they first hit public whitelist on virgo), and while they are a fun and interesting species, they are not without their problems. This PR seeks to resolve most of those problems, and rebalance some of their abilities that some may consider "powergaming". Below I have listed their issues, and how I have attempted to resolve them.

Problems with old prots / Feels bad to play as:
Having to blob for even the tiniest bit of damage.
Having to blob for replacing a limb of any kind.
Blobbing dropping everything you're wearing for the above, breaking rigsuits and some other things.

Gameplay that felt odd/Overpowered:
Proteans in blob form (simple mob) can smash open airlocks (even bolted), secure crates/lockers, and generally kill off most simple mobs so long as they have steel.
Full Refactor -> rebuild Basically grants an Aheal, trading 5 steel for removing ALL your damage after damage, even giving you a new NiF.
Blob form being able to be used as a quick escape from just about any situation.
Proteans could not be arrested, as they could just blob out of straitjackets/cuffs (prommies cant do it, why can prots)

New ability added:
Toggle Regeneration: (100 steel per operation, Regen Rate unchanged)
Going into blob form no longer instantly starts healing you, it must be toggled on/off with this ability.
Ability works while not blobbed too, but additionally costs power (nutrition) to run.
Ability will be turned off automatically if injured (excluding in blob form, cant quite figure that one out)
Ability will turn off automatically if you 1. Run out of steel. 2. Run out of power (blob form excluded) 3. Take damage (noted above)
Ability takes ~5 seconds to actually turn on, which you must remain still for (cant just tap regeneration on while youre in combat)
Ability -does- work if youre in crit, but simple mobs will guaranteed beat your ass if you try this.

Abilities Changed:
Blob form:
Has a 5 second Do after (that reports to nearby users that its occuring)
cannot be done if: Stunned (purple stun icon) Paralyzed (red stun icon) Or handcuffed (arrestable proots)

Mass adjust (size change):
Removed the steel requirement/refund when changing sizes.
Its more of a scene tool kind of thing, and none of the other size change mechanics require this silly resource.
It was also usable to just become smaller and gain more steel for regenning leading to some balance concerns.

Reform - Single Limb: 2000 steel (1 sheet)
Removed blobbing requirement, added a 10 second stay still period.
Fixed it taking your steel until -after- the operation is completed.

Reform - Whole body: 10000 Steel (5 sheets, all your capacity)
Swapped the old method out for a small recode of the "regenerate" verb
Has a 30 second stand still period.
No longer requires blobbing.
Will regenerate your Missing limbs and Fix/Regenerate internal organs (there's no way to do this otherwise)
Does NOT heal body damage, Leaving you with all your wounds you had previously (does fix stumps though and their associated damage)

Gameplay / Balancing changes (general):
Removed Blob form (simple mob's) damage:
No more smashing open Crates/Lockers Or opening airlock doors without access bolted or not (Looking at you alien ship)
Makes blob your "I need to get out of here, fast." Form, or shifts it towards scene play

You can now blow a proteans head off after 50 damage of any kind, no longer can they just tank hits here.
Groin/Pelvis also tagged as "vital" and damage here counts towards their death

The two above changes fix them recieving >200 damage then blobbing resulting in you just instantly gibbing

### DIFF
--- a/code/modules/mob/living/carbon/human/species/station/protean_vr/protean_species.dm
+++ b/code/modules/mob/living/carbon/human/species/station/protean_vr/protean_species.dm
@@ -31,7 +31,6 @@
 	breath_type = null
 	poison_type = null
 	//RS add (for healing ability)
-	var/heal_rate = 0.5 // Temp. Regen per tick.
 	var/prot_healing_allowed = FALSE	// Switches to FALSE if healing is not possible at all.
 	var/stored_brute = 0
 	var/stored_burn = 0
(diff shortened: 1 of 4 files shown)


TheDavestDave:
I feel like the blobbing out of cuffs is one of the reasons it's white listed. I also don't think blobbing to replace limbs is bad. I don't think removing damage from the blob isn't bad but I feel like resting should let you hide from mobs better or something. IDK it's not like we have a mob stealth system despite me dreaming we'll get  one.

that aside maybe sec should get like stablisiation cuffs or something for proteans and slimes. I think something that locks a shapeshifter in their current shape would be a fun scene tool if nothing else

crossexonar:
Addendum: Please only comment -after- having tested it, not on a theoretical. I'd rather this not devolve into a ' this feels like it might be X ' but a ' After reviewing the how and why, I have determined with testing that X is too powerful, or that Y feels weird. '

FartMaster69420:
A cohesion straight jacket could be made for prommies and proteans in a future PR if we wanna return the ability to escape cuffs, while giving sec a counter play. Could also block the use of the blob form when in a cohesion suit, allowing cuffs to work normally.

I don't play protean, but most of this seems pretty good QOL changes.

crossexonar:
Fixed a missing message when youre starting to blob, and added some of the emotes for blob form, it was great injustice that prommy blobs can *squish but proteans could not

Chargae:
Seems alright to me. Needed fixes applied and balancing things to the blob which I wasn't even aware were a thing.

code/modules/mob/living/carbon/human/species/station/protean_vr/protean_species.dm
```diff
@@ -39,25 +44,26 @@
...
-	heat_level_2 = 370 //Default 400
-	heat_level_3 = 600 //Default 1000
-
+	//RS Edit (why defining temp twice?)
+	cold_level_1 = -INFINITY
+	cold_level_2 = -INFINITY
+	cold_level_3 = -INFINITY
+	heat_level_1 = 420 //Proper temperatures
+	heat_level_2 = 480
+	heat_level_3 = 1100
```
Arokha, line 53: Good, toasty bois

code/modules/mob/living/carbon/human/species/station/protean_vr/protean_species.dm
```diff
@@ -30,7 +30,12 @@
 
 	breath_type = null
 	poison_type = null
-
+	//RS add (for healing ability)
+	var/heal_rate = 0.5 // Temp. Regen per tick.
```
Arokha, line 34: 'Temp'?

code/modules/mob/living/carbon/human/species/station/protean_vr/protean_species.dm
```diff
@@ -85,11 +91,12 @@
 		BP_R_FOOT = list("path" = /obj/item/organ/external/foot/right/unbreakable/nano)
 		)
 
-	heat_discomfort_strings = list("You feel too warm.")
+	heat_discomfort_strings = list("WARNING: Temperature exceeding acceptable thresholds!.")//Hot message more robotic
```
Arokha, line 94: I like it

code/modules/mob/living/carbon/human/species/station/protean_vr/protean_species.dm
```diff
@@ -137,7 +144,76 @@
 /datum/species/protean/handle_post_spawn(var/mob/living/carbon/human/H)
 	..()
 	H.synth_color = TRUE
+//rs Add (Healing ability non blob)
+/datum/species/protean/handle_environment_special(var/mob/living/carbon/human/H)
 
+	var/current_brute = TRUE
+	var/current_burn = TRUE
```
Arokha, line 151: Don't have these start out initialized to boolean and then replaced with some numbers down below.
You don't need to initialize them to anything if you don't want to use them yet, just do
```
var/current_brute
var/current_burn
```

I don't see a good reason to initialize them up here at all though.

code/modules/mob/living/carbon/human/species/station/protean_vr/protean_species.dm
```diff
@@ -137,7 +144,76 @@
...
+
+
+	// Heal remaining damage.
+	if(prot_healing_allowed)
+		//to_chat(H,"<span class='notice'> Healing Proc</span>")
+		if(H.getActualBruteLoss() || H.getActualFireLoss())
+			var/nutrition_cost = 0		// The total amount of nutrition drained every tick, when healing
+			var/nutrition_debt = 0		// Holder variable used to store previous damage values prior to healing for use in the nutrition_cost equation.
```
Arokha, line 159: Lots of indent, I suggest flipping these checks. It's the last thing in the proc, so you can just do...
```
// can't heal
if(!prot_healing_allowed)
  return
// no damage to heal
if(!H.getActualBruteLoss() && !H.getActualFireLoss())
  return
```

Then you can lose two layers of indent that is pushing this way over.

code/modules/mob/living/carbon/human/species/station/protean_vr/protean_species.dm
```diff
@@ -137,7 +144,76 @@
...
+					var/to_pay = 0
+					if(current_brute <= stored_brute)
+						nutrition_debt = current_brute
+						H.adjustBruteLoss(-1,include_robo = TRUE) //Modified by species resistances
+						stored_brute = current_brute
+
+						to_pay = nutrition_debt - current_brute
+
+						nutrition_cost += to_pay
+					else
```
Arokha, line 184: This is a loooot of stuff to put in 'handle_environment_special()'. Maybe move it to it's own proc that's just called from handle_environment_special, at least?

handle_passive_regen? or something, I dunno.

code/modules/mob/living/carbon/human/species/station/protean_vr/protean_species.dm
```diff
@@ -137,7 +144,76 @@
...
+						to_pay = nutrition_debt - current_burn
+						stored_burn = current_burn
+						nutrition_cost += to_pay
+					else
+						to_chat(H,"<span class='notice'> Damage Taken, Deactivating Regeneration.</span>")
+						prot_healing_allowed = FALSE
+						stored_brute = 0
+						stored_burn = 0
+						return
+					if(!refactory.use_stored_material(MAT_STEEL,100))
```
Arokha, line 203: I don't think it's possible, on THIS one, unlike the others (because there's no wait period between check and use) to exploit this.

But, you also just bail out of the entire handle_environment_special() proc here if they don't have enough metal which is sort of odd, and you never set prot_healing_allowed = FALSE again, so it's still true after bailing out so you'll just be back here again next life() tick. You also don't reset the stored values so that'll be broken at that point. But again not sure if it's possible to trigger this particular one, unlike the other cases.

Even so, it can be cleaner: I'd suggest you just have the last check up there (like you check prot_healing_allowed and if you have damage), the final thing before doing the healing (to ensure you don't pay metal without getting anything for it) is to just try and spend the 100 metal at the top. And if you do spend it (the use_stored_material() returns a truthy value), you go in and do all the healing.

So like at the top, you just check like...
prot_heal? no? return
have damage? no? prot_heal = false, reset numbers, and send message saying no damage to heal
have nutrition? no? prot_heal = false, reset numbers, and send message saying no power
spend metal? no? prot_heal = false, reset numbers, and send message saying no metal
and if you get past all those, just do the healing.

Just make sure you do the metal spend attempt last since it actually costs something and the rest are checks.
If you really want you can get_stored_material to check, but it's sort of pointless, you can save yourself a proc call by just 'trying to spend it' and use that as the check if you just do it as the LAST check.

Since if you do it like... try to spend metal, then check nutrition as an example, you have enough metal, it spends 100, then 'oops you don't have enough nutrition' and bails out after costing you 100 so just make it the LAST one. Okay I said that enough.

I'd make a proc that's just reset_passive_healing() and it unsets the allow healing var, and sets the two damage stored vars back to 0. Just because that's done in so many places, it would be nice as a proc.

You can even have it accept a reason that it prints to the user, so you don't need to put the to_chats everywhere, too.

Arokha (changes requested):

PR #432, opened 2024-07-10
Pet Carriers!

FartMaster69420:
Pet Carriers!

- Constructable from any autolathe!
- Can hold up to 3 tiny mobs (EG: mice), or 1 small mob (Catslug, monkey, doggo, ect)
- Can be locked to secure the critter inside and prevent escape.
- Locked critters can escape via resist to unlock the crate, and another resist to leave.
- The OOC Escape verb can also be used to instantly teleport out of a crate if you don't want to be there.
- People can be put in pet carriers! You can scoop anyone 20% smaller than yourself and put them in the pet carrier. Standard pickup prefs are respected for this interaction.

A bluespace pet carrier also exists, this is an advanced Pet Carrier that can hold larger mobs and places them into stasis when sealed (Hostile Mobs will not be able to attack while in stasis). The intent for these crates is to give xenobiology a way to transport hostile mobs. (Note, hostile mobs must be in a net to be trapped into a bluespace carrier.)

While the bluespace carrier exists, it is not currently buildable and is only accessible via Admin Spawn.

Ported from https://github.com/tgstation/tgstation/pull/33231

Sprites by WalterJe, detailed credits in the TG Station PR.


FartMaster69420:
Updated with attributions and licensing information as requested by @Lilly-Lira

lm40 (approved):

PR #437, opened 2024-07-13
Mail Update

Durandaal:
Updates the mail subsystem so only crewmembers on the manifest may be sent mail.


FartMaster69420:
Can talon still get mail? Because that would be funny I think

lm40 (approved):

PR #452, opened 2024-07-18
Protean refactor (reup)

crossexonar:
accidently deleted repository, original: https://github.com/TS-Rogue-Star/Rogue-Star/pull/429
Hello dear readers, I have played as a protean for a long time (Back since they first hit public whitelist on virgo), and while they are a fun and interesting species, they are not without their problems. This PR seeks to resolve most of those problems, and rebalance some of their abilities that some may consider "powergaming". Below I have listed their issues, and how I have attempted to resolve them.

Problems with old prots / Feels bad to play as:
Having to blob for even the tiniest bit of damage.
Having to blob for replacing a limb of any kind.
Blobbing dropping everything you're wearing for the above, breaking rigsuits and some other things.

Gameplay that felt odd/Overpowered:
Proteans in blob form (simple mob) can smash open airlocks (even bolted), secure crates/lockers, and generally kill off most simple mobs so long as they have steel.
Full Refactor -> rebuild Basically grants an Aheal, trading 5 steel for removing ALL your damage after damage, even giving you a new NiF.
Blob form being able to be used as a quick escape from just about any situation.
Proteans could not be arrested, as they could just blob out of straitjackets/cuffs (prommies cant do it, why can prots)

New ability added:
Toggle Regeneration: (100 steel per operation, Regen Rate unchanged)
Going into blob form no longer instantly starts healing you, it must be toggled on/off with this ability.
Ability works while not blobbed too, but additionally costs power (nutrition) to run.
Ability will be turned off automatically if injured (excluding in blob form, cant quite figure that one out)
Ability will turn off automatically if you 1. Run out of steel. 2. Run out of power (blob form excluded) 3. Take damage (noted above)
Ability takes ~5 seconds to actually turn on, which you must remain still for (cant just tap regeneration on while youre in combat)
Ability -does- work if youre in crit, but simple mobs will guaranteed beat your ass if you try this.

Abilities Changed:
Blob form:
Has a 5 second Do after (that reports to nearby users that its occuring)
cannot be done if: Stunned (purple stun icon) Paralyzed (red stun icon) Or handcuffed (arrestable proots)

Mass adjust (size change):
Removed the steel requirement/refund when changing sizes.
Its more of a scene tool kind of thing, and none of the other size change mechanics require this silly resource.
It was also usable to just become smaller and gain more steel for regenning leading to some balance concerns.

Reform - Single Limb: 2000 steel (1 sheet)
Removed blobbing requirement, added a 10 second stay still period.
Fixed it taking your steel until -after- the operation is completed.

Reform - Whole body: 10000 Steel (5 sheets, all your capacity)
Swapped the old method out for a small recode of the "regenerate" verb
Has a 30 second stand still period.
No longer requires blobbing.
Will regenerate your Missing limbs and Fix/Regenerate internal organs (there's no way to do this otherwise)
Does NOT heal body damage, Leaving you with all your wounds you had previously (does fix stumps though and their associated damage)

Gameplay / Balancing changes (general):
Removed Blob form (simple mob's) damage:
No more smashing open Crates/Lockers Or opening airlock doors without access bolted or not (Looking at you alien ship)
Makes blob your "I need to get out of here, fast." Form, or shifts it towards scene play

You can now blow a proteans head off after 50 damage of any kind, no longer can they just tank hits here.
Groin/Pelvis also tagged as "vital" and damage here counts towards their death

The two above changes fix them recieving >200 damage then blobbing resulting in you just instantly gibbing

+ Addressed issues @Arokha  Brought up

### DIFF
--- a/code/modules/mob/living/carbon/human/species/station/protean_vr/protean_powers.dm
+++ b/code/modules/mob/living/carbon/human/species/station/protean_vr/protean_powers.dm
@@ -321,6 +321,7 @@
 		return
 	else
 		//RS Edit (no blobbing while arrested/unconcious)
+		to_chat(src,"<span class='notice'>Your form starts to shift as you begin to collapse into a gooey blob.</span>")
 		visible_message("<b>[src.name]</b> starts to collapse into a gooey blob!")
 		if(do_after(src,10 SECONDS))
 			if(stat || paralysis || stunned || restrained()) //Double check to make sure we didnt get KO'd during our do_after()
--- a/code/modules/mob/living/carbon/human/species/station/protean_vr/protean_powers.dm
+++ b/code/modules/mob/living/carbon/human/species/station/protean_vr/protean_powers.dm
@@ -343,7 +343,7 @@
 					to_chat(src,"<span class='warning'>Blobbing interrupted.</span>")
 					active_regen = FALSE //Global Protean Ability active chec
 					return
-					nano_intoblob()
+				nano_intoblob()
 			active_regen = FALSE
 			//RS Edit End
 
--- a/code/modules/mob/living/carbon/human/species/station/protean_vr/protean_powers.dm
+++ b/code/modules/mob/living/carbon/human/species/station/protean_vr/protean_powers.dm
@@ -195,7 +195,7 @@
 					var/limb_path = organ_data["path"]
 					var/obj/item/organ/O = new limb_path(src)
 					organ_data["descriptor"] = O.name
-					to_chat(src, "<span class='notice'>You feel a slithering sensation as your [O.name] reform.</span>")
+					to_chat(src, "<span class='notice'>You feel a shifting sensation as your [O.name] reconsitutes.</span>")
 
 
 
--- a/code/modules/mob/living/carbon/human/species/station/protean_vr/protean_powers.dm
+++ b/code/modules/mob/living/carbon/human/species/station/protean_vr/protean_powers.dm
@@ -2,7 +2,7 @@
 ////
 //  One-part Refactor
 ////
-/mob/living/carbon/human/proc/nano_partswap()
+/mob/living/carbon/human/proc/nano_partswap()//RS Edit (Ability changed to not require blobbing)
 	set name = "Ref - Single Limb"
 	set desc = "Allows you to replace and reshape your limbs as you see fit."
 	set category = "Abilities"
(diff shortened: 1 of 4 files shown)


code/modules/mob/living/carbon/human/species/station/protean_vr/protean_powers.dm
```diff
@@ -267,19 +317,36 @@
...
 		to_chat(src,"<span class='warning'>You can only do this while not stunned.</span>")
 		return
 	else
-		nano_intoblob()
+		//RS Edit (no blobbing while arrested/unconcious)
+		if(active_regen == FALSE) // Prevents you from accidently blobbing while an ability is running
+			active_regen = TRUE;
+			to_chat(src,"<span class='notice'>Your form starts to shift as you begin to collapse into a gooey blob.</span>")
+			visible_message("<b>[src.name]</b> starts to collapse into a gooey blob!")
+			if(do_after(src,10 SECONDS))
```
Arokha, line 341: Not a big fan of the new 10s delay on blob form

code/modules/mob/living/carbon/human/species/station/protean_vr/protean_powers.dm
```diff
@@ -141,9 +155,13 @@
 		to_chat(src, "<span class='danger'>Remain still while the process takes place! It will take 5 seconds.</span>")
 		visible_message("<B>[src]</B>'s form collapses into an amorphous blob of black ichor...")
 
-		var/mob/living/simple_mob/protean_blob/blob = nano_intoblob()
-		active_regen = TRUE
-		if(do_after(blob,5 SECONDS))
+		//var/mob/living/simple_mob/protean_blob/blob = nano_intoblob() //RS Edit (Ability changed to not require blobbing)
+		active_regen = TRUE //Global Protean Ability active check
+		if(do_after(src,5 SECONDS))
+			if(refactory.get_stored_material(MAT_STEEL) < refactory.max_storage) //RS Edit changed to not take steel untill operation completed
```
Arokha, line 161: You migrated the others to use_stored_material at the top but left this one as get_stored_material

Arokha (dismissed):

code/modules/mob/living/carbon/human/species/station/protean_vr/protean_powers.dm
crossexonar: Replaced with a Confirm Popup text (can just press enter after clicking)

code/modules/mob/living/carbon/human/species/station/protean_vr/protean_powers.dm
crossexonar: Should be fixed now, had to move things around a little bit though.

crossexonar:
Had to add an integer check to blob heal proc, it was looping as active when it shouldnt've been.

Arokha (approved):
Probably fine

PR #453, opened 2024-07-18
Port chomp crawling (reup)

crossexonar:
Original: https://github.com/TS-Rogue-Star/Rogue-Star/pull/451

Ports chompstation crawling - https://github.com/CHOMPStation2/CHOMPStation2/pull/7822
With these fixes:
https://github.com/CHOMPStation2/CHOMPStation2/pull/8154
https://github.com/CHOMPStation2/CHOMPStation2/pull/8047
https://github.com/CHOMPStation2/CHOMPStation2/pull/7900

(and added a stun(3) to simple mob vore sequence)

Concerns with current implementation:
Rest crawling to bypass tables feels cheaty compared to dragging your sprite atop of, but I'm leaving that up to project leads
(its easy to turn it off)

### DIFF
--- a/code/_onclick/click.dm
+++ b/code/_onclick/click.dm
@@ -68,7 +68,7 @@
 		CtrlClickOn(A)
 		return 1
 
-	if(stat || paralysis || stunned || weakened)
+	if(stat || paralysis || stunned) //CHOMPedit, removed weakened to allow item use while crawling
 		return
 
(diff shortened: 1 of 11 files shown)


SpringSkipper:
I will say that crawling to get over tables is much faster than the existing system, but also the existing system is _obnoxiously_ slow and just encourages *flip spam.

lm40 (approved):

PR #455, opened 2024-07-19
Adds cargo oties! 

crossexonar:
Adds cargo oties to join as mob. (sprites of oties as always by https://github.com/Verkister Provided with Creative Commons license)
Gives security access to sec oties
Gives cargo access to cargo oties.

Based on same logic as gatslug and technician nermley (rogue star scugs)

Helpful if they actually have department access, but understandable if access is not wanted!


Lilly-Lira (commented):
Reviewed!  No concerns with code or attribution on my end,

lm40 (approved):

PR #500, opened 2024-08-03
Revert "Hair whitelist removal"

Very-Soft:
Per Lira request


Lilly-Lira:
Hair PR was merged before it was ready for prime time.  But we can just do another PR to finish cleaning it up.

PR #524, opened 2024-08-10
Tweak vchat winset

Arokha:
Untested, also adds a debug proc called client_winset since it might be useful in testing if this doesn't correct the behavior.


Lilly-Lira:
Hard to tell if this fixes it since I'm not sure I get the issue when running the game off my computer, but doesn't seem to cause any stability issues.  Seems ready to merge to see if this broadly fixes the refresh bug.

lm40 (approved):

PR #536, opened 2024-08-17
Makes absorbs escapable and other virgo ports

Screemonster:
Various ports of stuff I've done on virgo

Adds ability to struggle free when absorbed on a %chance basis - it's necessary to succeed a normal escape roll first and the absorb escape roll is made _after_ the escape timer to see

defaults mobs to 20% chance

adds ability to autotransfer to a new belly on absorb (so for instance you can send them to one with a high starting escape chance but low unabsorb chance)

adds selective mode chance alongside absorb and digest

tweaks gaslamps and succubi to use these

note: these new features are NOT currently editable on the vore panel and the default values are such that nothing will change


Screemonster:
a port of https://github.com/VOREStation/VOREStation/pull/15559 will have to be made for these things to be properly customisable but this touches the vore panel and save files and I don't know what I'm doing with either of those things

succubi messages that rely upon that PR are present in this one but commented out

PR #542, opened 2024-08-19
Update virgo_minitest-1.dmm

Michab02:
adds chem stuff because Sari asked


Very-Soft:
I am not wild about this because, Minitest is meant in large part to be like, minimum stuff to be able to test whatever feature you're working on. A large point of using it is that it loads in like 5 seconds. If we keep adding stuff to it for every individual test that people might want to do, then it's going to take longer and longer to load.

If someone needs to test something from chemistry, it's not all that hard to open up the map editor, plonk down the machines from chemistry, and then compile it for their individual test.

PR #556, opened 2024-08-30
removes river leeches from pet spawners

Screemonster:
removes leeches from "all theme" and "sif" pet spawner pools

leaves leeches in the global sif list just in case it's used somewhere that leeches are actually wanted


Arokha (approved):
Seems okay. Usually I don't like list-modification in this way and would suggest against it on anything that runs more than once, since it just iterates through the list until it finds this thing. If it ran more than once, I'd say make a new list you can reuse places. 
But, these run once, so it's probably fine.

Screemonster:
I can do that instead if you prefer, only reason I did it this way was so it'd inherit changes if the base sif list ever got changed for whatever reason

PR #594, opened 2024-09-17
[READY] Adds Synthetic Digitigrade Legs

Durandaal:
Adds synthetic digitigrade legs! This pulls from CHOMPStation's implementation of digitigrade synthetic legs, [PR 5565](https://github.com/CHOMPStation2/CHOMPStation2/pull/5565). The only prosthetics that will work with this implementation is the DSI prosthetics, as they are indistinguishable from normal limbs.

Additionally, proteans can also use teshari DSI prosthetics.

~~- TODO: Fix how synthetic digitigrade legs are colored.~~

~~The current problem with this PR is that, it seems like synthetic Digitigrade colors are applied twice, and I'm not quite certain where or how as of yet due to my lack of experience. I'll continue to poke at this, but for now I'm putting this up as a draft PR to see if someone else with more experience has a better idea.~~

PR is ready. Found a logic issue in one of the other functions in the proc. I'm not sure if it's an elegant solution, but it's at least an effective merger between the Modular Chomp files and our forked version of VOREStation's code.

### DIFF
--- a/code/modules/organs/organ_external.dm
+++ b/code/modules/organs/organ_external.dm
@@ -27,6 +27,7 @@
 	var/spread_dam = 0
 	var/thick_skin = 0                 // If a needle has a chance to fail to penetrate.
 
+
 	// Appearance vars.
 	var/nonsolid                       // Snowflake warning, reee. Used for slime limbs.
 	var/transparent                    // As above, so below. Used for transparent limbs.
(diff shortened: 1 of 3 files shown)


Durandaal:
Issues are resolved, ready for merge.

SigholtStarsong:
SPEEDMERGE, SPEEDMERGE, SPEEDMERGE

PR #596, opened 2024-09-20
Adds new ability to shadekin!

ZayahPapaya:
This adds a new "Phase Flicker" ability to shadekin by default. To use, you must be adjacent to one or more lights while in phase, then simply use the verb / button. It will flicker a random light within range a few times. Very spooky, very cute!


amenitykitten:
Very low CD.
No cost.

Gonna go ahead and voice my concerns here because I've already voiced my concerns on the Discord about rapid flickering/flashing effects.

This needs a disability filter.

Very-Soft (changes requested):
RS edit code comments required

PR #618, opened 2024-10-08
Prevents Pickup of Self-Occupied Items

Durandaal:
[Ports VOREStation PR15780](https://github.com/VOREStation/VOREStation/pull/15780), which prevents someone who occupies an item from picking it up (ie. a micro inside of food).


FartMaster69420:
Aww, that’s my favorite bug.

PR #624, opened 2024-10-12
Antler Sprite Fix

Lilly-Lira:
Found a typo while testing another PR


probbrie:
damn, beat me to it (didn't want to include in the original cuz atomization), no idea where the typo originated but it's been in every codebase that has them.

PR #630, opened 2024-10-17
Swaps machines in Talon mining

SigholtStarsong:
What it says on the tin. Goes from lossy to just persistent.


crossexonar:
No, talon must SUFFER forever.
(jk I've been wanting this too)

PR #646, opened 2024-11-05
Adds Mind Binder device from Chompstation

Cameron653:
Original PR: [Here](https://github.com/CHOMPStation2/CHOMPStation2/pull/7569). Credit to [Fracshun](https://github.com/Fracshun) for making the Mind Binder. Cryo-Crash bugfix credit [here](https://github.com/CHOMPStation2/CHOMPStation2/pull/7673)

### Adds a 'Mind Binder' device, based off the Sleevemate
- Has two toggle modes: Other and Self
- Other allows you to download a mind from someone OR pull a mind out of an already possessed object OR put an already downloaded mind into an object.
- Self allows you to put your mind into a body that has no mind (ex: freshly made sleeve) OR an object.
- Adds it to R&D as an item that requires Uranium and Illegal tech.
- Has a bajillion warnings so you don't accidentally bind yourself into an item or person unknowingly without realizing it could be a VERY BAD IDEA

### Adds various small fixes that were added with the Mind Binder
- Makes possessed objects no longer crash the game if someone goes into the cryo with them
- Makes possessed objects no longer send the person possessing said object go to the void if they're gurgled
- Makes it so when plushies are renamed, the mind is renamed as well.

#### The Chompstation version has it so mobs can be possessed. This version does not have that.
- Given the fact that going through and touching every file in the game so that mob possession is possible via the Mind Binder would take ages, I am going to leave that feature to either my future self OR whoever wants to tackle that task

### DIFF
--- a/code/game/machinery/cryopod.dm
+++ b/code/game/machinery/cryopod.dm
@@ -438,6 +438,11 @@
 	items -= announce // or the autosay radio.
 
 	for(var/obj/item/W in items)
+		if(islist(W.possessed_voice)) //RS Edit Start
+			for(var/mob/living/V in W.possessed_voice)
+				despawn_occupant(V)
+			W.forceMove(get_turf(src)) //this crashes the MC, so now they get spat back out.
+			continue //RS Edit End
 		//VOREStation Addition Start
 		if(istype(W, /obj/item/device/pda))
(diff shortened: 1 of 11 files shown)


lm40:
I have been dead set against any TF mechanics forever and this basically is just TF mechanics with an extra step to me.

PR #649, opened 2024-11-10
Adds Give Reagents, and Purge All Reagents as Smite Options

Durandaal:
Adds two new smite options, "Give Reagents" and "Purge All Reagents" to hopefully help things go a little bit smoother as opposed to hunting for view variables.

Give reagents allows an admin to apply any reagent to any valid location (Stomach, bloodstream, skin), up to five units. 

Purge all reagents does exactly as it says, it purges the target of all reagents. 

This PR borrows code from [VOREStation PR 16414](https://github.com/VOREStation/VOREStation/pull/16414), though has been slightly adapted for our use case in the smite panel as opposed to the admin effects panel they have there.


FartMaster69420:
Aww, only 5 units? What if an admin needs to turn someone into a maxcap?

PR #654, opened 2024-11-12
20241111_1955_KH Unwhitelist Akhani.

lm40:


code/modules/mob/language/station_vr.dm
```diff
@@ -159,9 +159,11 @@
 	flags = 0
 /datum/language/human
 	flags = 0
+/datum/language/tajaranakhani
```
Lilly-Lira, line 162: Needs //RS Add comments

Lilly-Lira (changes requested):
Seems fine, just needs a code comment

PR #655, opened 2024-11-13
Protean Regen Fix

Cameron653:
- Fixes metal from being immediately used up as soon as you attempt the protean regen ability, only if you complete the ability. (Sari-N, lines 190, 199-200)

- Fixes 'remain still' text to only appear after the checks to see if you can perform it are done, so it doesn't say you're doing it immediately to be followed up by denial text.

- Removes duplicate text. (They were showing up twice in game while bugtesting.)


crossexonar:
I'm sari bork, and I approve this message.

lm40 (approved):

PR #657, opened 2024-11-13
Adds 'Natural Artist' trait

Cameron653:
Does what it says on the tin.

### Adds a neutral 'Natural Artist' trait for individuals that are made of goo, paint, nanites, or have some other natural, innate ability to paint.

Has two new buttons:
- Adjust Artistic Color (Allows you to select what color you want)
- Conjure Natural Brush (Allows you to conjure/retract your paintbrush)

Did a bunch of kind-of-somewhat-extra stuff after consulting with the God (aka, old code) to make sure you couldn't do cheaty things with it.
- Throwing your appendage (OW) makes it immediately despawn. 
- You can't place your appendage in a backpack. Or anywhere for that matter.

I had originally wanted it so you could just click on the canvas so you could paint (without an in-hand brush) but realized after implementing it that while I had done the BYOND code side complete, I had failed to realize that it also required touching TGUI and revamping how it handles. 
Which, given TGUI code is not my forte, I chose to go a different route (this PR with the 'brush appendage')

### DIFF
--- a/code/game/objects/structures/artstuff_rs.dm
+++ b/code/game/objects/structures/artstuff_rs.dm
@@ -13,7 +13,6 @@
 
 /obj/item/paint_brush/organic/Initialize(location)
 	..()
-	START_PROCESSING(SSobj, src)
 	if(ismob(loc))
 		visible_message("[loc.name] pulls out an organic paintbrush of some sort!")
 		creator = loc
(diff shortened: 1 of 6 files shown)


Lilly-Lira (approved):
Looks good to me :3

PR #662, opened 2024-11-16
Smooches

Cameron653:
- Allows to use *blowkiss to blow a  kiss at someone!

### Features:
- New emote: "blowkiss" that allows you to blow a kiss at  a target. 
- It chases them down for 7 tiles and if it collides with them, it says 'the kiss leaves a smooch mark on [target]'s cheek!'
- Has a toggle under preferences to turn smooches on/off. So if you don't want to be targeted, you can turn them off!
- Has an audible toggle as well. No sounds currently, but if someone did want to add sounds to it, I put all the framework there!


Screemonster:
okay but now my brain's going to "what if this but you could use it for vore"

homing missile that sends a pair of lips chasing after someone and if it catches them then gulp

I guess this is more of a wizard spell kinda deal

Arokha:
I see you added toggles for sounds and stuff, but didn't add any sounds?

PR #667, opened 2024-11-18
Waddler / Animated Movement (Modified Bubberstation Trait)

Cameron653:
I searched as far back as I could to find credit for waddling. See: https://github.com/Bubberstation/Bubberstation/commit/14f89968acca662c3886257c93cdf689013ca861 Commit 14f8996 by vuonojenmustaturska on Aug 23, 2018 is as far as I could find.

This is a heavily modified version of the waddler trait.
DEFAULT/RECOMMENDED:

MAX Settings:

Features:
- Adds waddling / Animated Movement! This is a neutral trait. (Borgs get it by default but have a toggle to turn it on/off!)

- See videos above for the 'default' (suggested) waddle and the 'max' waddle

- Makes it simple to proc_call onto someone to allow them to manually adjust their waddle variables! Need someone to JUMP AROUND during an event? Look no further!

- Added (and commented out) a bunch of various ways to further adjust the system. Want it to be a preference? It's able to be commented in! Want to make it so drugs or statuses adjust it? Just have to uncomment one section.

### DIFF
--- a/code/game/Rogue Star/waddle.dm
+++ b/code/game/Rogue Star/waddle.dm
@@ -74,11 +74,11 @@
 		return
 	waddle_z = Z_height
 
-	var/min = tgui_input_number(usr, "Put the desired waddle backwards lean. (-4 is default. -12 min, 0 max)", "Set Back Lean", -4, 0, -12)
-	if(min < -12 || min > 0 )
+	var/min = tgui_input_number(usr, "Put the desired waddle backwards lean. (4 is default. 0 min, 12 max)", "Set Back Lean", 4, 12, 0)
+	if(min > 12 || min < 0 )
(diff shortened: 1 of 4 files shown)


Zoldorf:
Don't add a toned down version, it really should be the full on proper waddle or nothing, because this version doesn't look like a waddle, it looks like you're just randomly stumbling slightly.

It's a feature that isn't supposed to be overly serious to begin with, it'd be for gremlin characters and lil' guys.

It should also be noted that it is a trait that can be taken or ignored and not something every mob has.

Don't wanna waddle? Don't take the trait.

CEO-Of-Dogborg-Corp:
How do you turn it off as borg.

PR #670, opened 2024-11-23
Journalist Time

Lilly-Lira:
Adds a Centcom Journal Role!


lm40 (changes requested):
Probably the /datum/job/centcom_officer/get_access() should be /datum/job/centcom_journalist/get_access()

lm40:
Or actually on second thought maybe make this whole thing a subtype of centcom_officer unless you have some compelling reason it should be different

Lilly-Lira:
It is efficient now :3

lm40 (approved):

PR #683, opened 2024-11-27
Shadekin Light Flicker adjustments

Cameron653:
### Makes it so flashlights can flicker
- Reworks the shadekin phase-in proc. Makes it so the flicker length, light break %, and the light flicker color is adjusted!

Features: When shadekin use the toggle, it lets them set three things
- How long the light will flicker for in seconds (10 to 15 seconds)
- % chance of the lights breaking upon phasing in (Used to be hardcoded to 0. Now can be set from 0-10%). Can be admin VV'd to allow event shadekin to have higher break chances if desired.
- Allows shadekin to set the color they wish for lights to flicker as when they phase in!
- Makes it so shadekin warping in will make flashlights flicker the selected color as well!

Side things:
- Allows the flicker() proc to allow an input for what color to flicker as.
- Allows the flicker() proc to set the color of the light to what it originally was before it was changed if there is a new color it's flickering to.

Examples of randomly selected colors:

Example of flashlights flickering: (The first example doesn't have the updated 'more colorized' flashlight version. See the second one for up to date example. It's just to show it works on held flashlights.)

### DIFF
--- a/code/game/objects/items/devices/flashlight.dm
+++ b/code/game/objects/items/devices/flashlight.dm
@@ -501,15 +501,18 @@
 	spawn(0)
 		var/original_color = light_color
 		var/original_on = on
+		var/datum/component/overlay_lighting/OL = GetComponent(/datum/component/overlay_lighting) //BEWARE, ESOTERIC BULLSHIT HERE.
 		for(var/i = 0; i < amount; i++)
 			if(flicker_color && light_color != flicker_color)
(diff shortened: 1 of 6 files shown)


SigholtStarsong:
Probably outside the scope of this project, but would it be possible to make flicker() and such reset the light to whatever color it already was? 

During the Redspace Blowout events, Shadekin phasing reset lights to white, ruining the red glow of the station.

PR #714, opened 2024-12-16
Borg Multibelly, Glowing stomachs.

Cameron-The-Raven:
So, this PR was ORIGINALLY just to fix the vore panel for borgs and add multi-stomach capability for borgs (much like how humanoids have taur tummy+stomach), but the soon-to-come PR quickly became more and more complex, requiring:

Additions:
- A complete rework of borg overlay code  (See: #708)

- Multi-stomach capability (This PR) (Ported from Chompstation, modified to work here!)

- Glowing stomachs (This PR) (Original code)

- TOUCHING TGUI 
(THIS TOOK AN ENTIRE DAY TO ADD ONE BUTTON. ONE. BUTTON.) (Also this PR. Original code (screw TGUI))

Unintended Additions:
- Humanoids are now able to enjoy glowing stomachs as well. I realized how ANNOYING it would be to make it cyborg only in TGUI (Attempting to do such resulted with a LOT of problems, which led me to a massive waste of time) so it's just a preference now! You can toggle it on and off at will in case you don't want glowy bellies for a scene or something like that.
This also applies to taur bellies!

Changes:
- Gives borgs the 'glow body' trait because if a borg wants to have a passive glow (they have red accents and want to glow red) it allows them to do such without having their MASSIVELY BRIGHT LIGHTS ON. Also means if someone has a glowing belly they can set their light to be the same color as their glowing belly for FLAIR. 
(This doesn't use energy because it's meant to be a passive thing and borgs having a pda-level of light REALLY does not matter)

Fixes:
- Makes clownborgs spawn with the 'sleeper' stomach (This was missed during bugtesting as they only have one sprite. Borgs with only one sprite didn't call update_multibelly() resulting in their stomachs never being initalized.)
- Makes borgs (and mobs) able to use the vorepanel!
- While talking to a friend, I accidentally stumbled across venomous bite's 'Chemical Refresher' option being bugged...I fixed it while I was here since it was a 1 character fix. (If this is wanted to be moved to its own PR I can, but...I don't imagine this being an issue.)

Notes:
- Adds documentation and instructions on how to properly write your .dmis so spriters can spend more time spriting and less time reading through code trying to figure out what they CAN and CAN NOT do.

Glowing tummy photos:
image
image
image
image
image

image

dreamseeker_2024-12-16_23-40-08

# TGUI WILL HAVE TO BE RECOMPILED VIA (bin/tgui ) AFTER THIS PR IS MERGED


amenitykitten:
**AY YO IS IT NORMAL FOR MY FOOD TO GLOW?**

PR #720, opened 2024-12-26
Adds an alternate Talon Captain title

Cameron-The-Raven:
Adds alternate Talon Captain title: Rogue Trader


crossexonar:
truly rogue. if only their other roles had silly alt titles

PR #730, opened 2025-01-08
Hypospray Mk1 Rework

crossexonar:
Text to be filled out, Testing the PR to make sure I didnt mess anything up.


crossexonar:
HOKAY, this is kind of a big one.
Replaces the original Mk1 Hypo-Sprays with an upgraded one.
Does NOT replace the Mk2 Hypo-Sprays by pooj.

Adds these new hypos to the Nanomed, CMO locker, and replaces the prototype hypo-spray printable in science

These were originally meant to be a midground between the Mk1 and Mk2's but ended up evolving.

Uses normal vials from viro for the basic mk1 Hypospray.
CMO version can use both vials and bottles.

Made the chem-master able to PRINT vials directly now.

Features of hypo's:
Can click on the hypo with a vial to swap directly to that one.
Can click the hypo onto a vial to swap directly to that one.
The hypokit's can be worn as belts to dedicate yourself to the way of the chem (a tiny bit more space efficient than using a belt + vials or belt + bottles with the CMO one)

Injection speed is based on how much of a chem you are injecting, lower amounts are faster (with a minimum time) and higher amounts are longer.

Implimented the mk1 behaviour of Resisting injection.

Can set the injection rate on the vials / bottles, or over-ride it on the hypo itself.

todo behavior: Make the hypo's slower when being used on somone wearing a voidsuit, similar to a syringe.
Make a printable upgrade that bypasses the above, and adds instant injection if 5u or lower (unless resisting the injection like the mk1) ((basicly mk1 behaviour when upgraded.))

crossexonar:
That SHOULD be everything for the initial testing/upload, I'll be writing the upgrade system later on after we make sure this change doesnt set medical on fire, Removing the draft status for now.

FartMaster69420:
Is there a reason we're calling this Mk1s? If they're better than the Mk2s I think that's just going to lead to confusion for new players.

crossexonar:
> Is there a reason we're calling this Mk1s? If they're better than the Mk2s I think that's just going to lead to confusion for new players.

Well it directly replaces the Mk1 hyposprays, and has limitations to what it can and cant do.

I see the mk2's as more of a "mass produced model" you get more of them, theyre marketed to the basic consumer with lots of pre-made packs (brute burn etc)

But if you have any name suggestions, feel free, im sure something will stick.

crossexonar:
Renamed them to iHypo by WolfApple, as they are white and I found it funny.
added in some requested functions by kenzie.
gave the hypokits 60u of the basic chems so that people dont just cannibalize the mk2's in the vendors for their chems.  (I'm sorry chemist mains)

Zoldorf:
I second Cameron.

keane0tts:
It's not secret that I'm a fan of hyposprays (I'm a Trekie after all), and so I'm a huge fan of this change.

I do agree with Cameron on the medicine issue. While I think it's fine to have SOME chems laying around in case of an emergency, any doctor answering a fax should be able to use the medkits in storage to stabilize (hell, even completely cure) a patient when answering a fax. 

Just having these new hypos available to be picked up and the ability to print vials from the chemmaster should be enough, while also keeping chemists relevant.

keane0tts:
It's not secret that I'm a fan of hyposprays (I'm a Trekie after all), and so I'm a huge fan of this change.

I do agree with Cameron on the medicine issue. While I think it's fine to have SOME chems laying around in case of an emergency, any doctor answering a fax should be able to use the medkits in storage to stabilize (hell, even completely cure) a patient when answering a fax. I don't like the existing hypospray kits in the medivends, and I don't think the new hyposprays should come with chems either. 

Just having these new hypos available to be picked up from the medivend and the ability to print vials from the chemmaster should be enough, while also keeping chemists relevant.

FartMaster69420:
I agree with what others have said here, keeping Mk2s is redundant.

If we absolutely have to keep them they should be moved out of the health vendors to reduce clutter.

keane0tts:
Oh shit, forgot to add that I personally think the white/blue ones should be the base model and the black/gold ones should be the upgraded model. Personally, I prefer medical tools not looking like guns.

crossexonar:
> Oh shit, forgot to add that I personally think the white/blue ones should be the base model and the black/gold ones should be the upgraded model. Personally, I prefer medical tools not looking like guns.

They are, dork.

crossexonar:
Would something like this address everyones concerns about chem bloat?
image

or having them put down near the other first aid kits and move the ml3m + new hypos up near the door 
(im not sure about the rascals pass map would have to look)

crossexonar:
better pictures of what it would look like on the two maps:
image

image

LanderExqueak:
It works for me, seems like a good place to have the 'Oh s#!t' chems for when someone hasn't had a chance to make any, and reduces bloat in the medivends.

keane0tts:
Like I said before, the medkits have everything you need to stabilize or even cure a patient if you're answering a fax and have to act in the moment. My preference would be just getting rid of them, but I know that's not going to happen. No harm in voicing my opinion about it anyway.

This is a fine compromise, though. Fewer kits overall, so I like it. I approve!

lm40:
I'll try to wrap this up this week or at worst by the close of the month I think it is fine I just have to do a final test/review at some point

lm40 (approved):
lgtm

PR #743, opened 2025-01-16
20250115_1913_KH Fixes people flipping back and forth while pinned.

lm40:
This definitely fixes the issue in question but DNM until I can be relatively sure it doesn't break the feature it is patching.


Durandaal:
This seems to resolve the issue, and does not discernibly impact the scaling-switch mode.

PR #762, opened 2025-02-03
Add a spont vore interaction for buckling

TheGreatKitsune:
Adds in a new spont vore type, buckling yourself or getting buckled to something that already has something buckled to it will do a stumble-vore like situation themed as sitting on someone.


crossexonar:
is this seat taken?

PR #765, opened 2025-02-05
Shadekin phase Fixes Pt.1

crossexonar:
Ported from the work of the legendary GUTI from chompstation.

fix: Step messages when stepping over a smaller, phased shadekin
fix: Shadekin catching items when phased
fix: Thrown item displaying messages and harming phased shadekin
fix: Slashing affecting phased shadekins
fix: Listening in, licking and smelling picking phased shadekin as targets
fix: Vines buckling phased shadekin
fix: Rogue vending machines targeting phased shadekin
fix: Mobs attacking phased shadekin
fix: Mobs eating laying down phased shadekin
fix: Thermals exposing phased shadekin
fix: Water and lava affecting phased shadekin
fix: Moving from space to station tile affecting phased shadekin
fix: Shadekin sliping on wet tiles
fix: Targeting a gun to a shadekin doesn't break when they phase out
fix: Phased shadekin squishing cockroaches
fix: Phased shadekin hallucinating if they stare at the SM

Future fixes in another PR: stopping kin from getting zapped by tesla, and probably other things awu.

https://github.com/CHOMPStation2/CHOMPStation2/pull/7366
https://github.com/CHOMPStation2/CHOMPStation2/pull/7484
https://github.com/CHOMPStation2/CHOMPStation2/pull/8267

### DIFF
--- a/code/__defines/misc.dm
+++ b/code/__defines/misc.dm
@@ -8,6 +8,7 @@
 #define INVISIBILITY_LIGHTING             20
 #define INVISIBILITY_LEVEL_ONE            35
 #define INVISIBILITY_LEVEL_TWO            45
+#define INVISIBILITY_SHADEKIN			  55 //RS add Chomp port #7484
 #define INVISIBILITY_OBSERVER             60
 #define INVISIBILITY_EYE		          61
 
--- a/code/game/Rogue Star/_HELPERS/mobs.dm
+++ b/code/game/Rogue Star/_HELPERS/mobs.dm
@@ -1,0 +1,14 @@
+/atom/proc/living_mobs_in_view(var/range = world.view, var/count_held = FALSE)
+	var/list/viewers = oviewers(src, range)
+	if(count_held)
+		viewers = viewers(src,range)
+	var/list/living = list()
+	for(var/mob/living/L in viewers)
+		if(L.is_incorporeal())
+			continue
+		living += L
+		if(count_held)
+			for(var/obj/item/weapon/holder/H in L.contents)
+				if(istype(H.held_mob, /mob/living))
+					living += H.held_mob
+	return living
--- a/code/game/atoms_movable.dm
+++ b/code/game/atoms_movable.dm
@@ -413,6 +413,7 @@
 			if(A == src) continue
 			if(istype(A,/mob/living))
 				if(A:lying) continue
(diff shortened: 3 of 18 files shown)


ZayahPapaya:
If you're doing future fixes for shadekin, could you have a look at the interaction of Body Glow + Create Shade? It completely bricks glow rather frequently.

Kashargul:
This opens up a lot of exploits, just to let you know.

crossexonar:
> This opens up a lot of exploits, just to let you know. https://github.com/CHOMPStation2/CHOMPStation2/pull/9786/files
> 
> follow up for several issues that will turn interactions one sided or allow weird funnies up to MC crashes. You maybe want to double check about the vore interactions though as that's more a codebase depending thing.

Could you please detail what specifically causes the MC crash? most of what I see in the changed files are related to throw mechanics while the rest seems to be balance decisions (which I am not touching with this SPECIFIC pr, but will be doing so after discussing with headmins.)

Kashargul:
> > This opens up a lot of exploits, just to let you know. https://github.com/CHOMPStation2/CHOMPStation2/pull/9786/files
> > follow up for several issues that will turn interactions one sided or allow weird funnies up to MC crashes. You maybe want to double check about the vore interactions though as that's more a codebase depending thing.
> 
> Could you please detail what specifically causes the MC crash? most of what I see in the changed files are related to throw mechanics while the rest seems to be balance decisions (which I am not touching with this SPECIFIC pr, but will be doing so separately after discussing with headmins.)

It's the bump check, if you bump into some things, I didn't bother to dig it much further back in the day as the entire interaction led to other issues as well like phase blocking doors in attack stance etc.

Solazius:
If I'm not mistaken, shadekin in phase can already block doors on the live server right now. That has nothing to do with this PR as far as I am aware.

ZayahPapaya:
Can confirm they can already block halls and keep airlocks from closing while in phase in harm intents.

Kashargul:
> If I'm not mistaken, shadekin in phase can already block doors on the live server right now. That has nothing to do with this PR as far as I am aware.

The blocking is an ancient bug and as said it was the reason why we didn't dig any further into the MC crash and just removed all of the bump checks while phased. It as more like, why not get rid of the blocking and some other bump things as well instead of digging for days to find the exact line causing the issue.

PR #766, opened 2025-02-05
Adds Refreshed Talon Uniforms

ChromeFoxxity:
**Updated Icons**
image

**Updated Captain**
image

**Updated Pilot**
image

*Yes, I know the pilot is wearing a harness, but people can take it off just as easily, and I didn't bother showing them all because that's a lot of work.*

Basically adds a few retextured items to the loadout for Talon members. Maybe editing a few talon members' shoes. Because sneakers for some reason just didn't fit them.

Some of these icons are not my own. They are specifically from TG Station, and are retextured, heavily edited variants.


code/modules/clothing/clothing_overrides.dm
Very-Soft, line 1: This should probably have a _rs suffix, ot a comment in the file denoting it as an RS file

ChromeFoxxity: Done.

PR #768, opened 2025-02-08
Minor Talon Fixes

ExabyteDiscord:
Fixes both of the port and starboard cargo hatches to not open both airlocks when trying to cycle out due to sensors being placed inside of a door, as well as fixes the privacy door lock for the Talon Pilot's room.


ExabyteDiscord:
👍

PR #776, opened 2025-02-12
Shadekin Tweaks

Very-Soft:
Fixed some unintended behavior and adjusted the balance of a few things.

Of note, when a shadekin is phased out, as far as the lore is concerned, they are moving into an alternative plane of reality. Because SS13 is a videogame, they effectively just become invisible, but ICly they're [somewhere else entirely], but able to peek in from the outside to determine where they want to end up if they phase back across. The changes that have been made are in service to that fantasy, and in general ought to make things slightly more fair for the crew.

Shooting guns and dropping grenades and giving security a play by play of event antagonist activities where they can't get you was never the way they were intended to be played. This should bring some of that more into line.

-Stopped ClickOn from working for incorporeal mobs. This stops shadekin from being able to shoot and use items on people while phased out. This does stop certain keybindings such as shift click from working while phased out, but you can always right click, or use the verb if you need to examine someone.

-Stopped activate held object from working for incorporeal mobs. This stops shadekin from being able to activate items in their hands while phased out, such as grenades.

-Stopped drop from working for incorporeal mobs. This prevents shadekin from being able to drop objects while phased out.

-Phase shift now EMPs the user, temporarily disabling or damaging devices they may be using. This will stop radios from broadcasting, and should hang up communicator calls. This uses EMP 5, which is not designated for damage to silicons or organ damage. Added an extra consideration for EMP 5 for pAIs, so that this level of EMP is not actually dangerous to them (as any EMP at all has a significant chance to instantly kill them.)

-Prevents the use of the TGUI panels for PDA, radio, and communicators while the user is phased out. This will prevent shadekin from activating these devices and sending covert messages out

-Prevents phased out shadekin from speaking into the radio.

-Prevents phased out shadekin from pulling objects (they can still pull mobs), to prevent the use of any such objects passive effects while phased out, such as radios with the mic broadcasting

-Adjusted shadekin energy gain values. All non-yellow-eyed shadekin will no longer lose energy in the light. As a result of this, reduced some of the energy gain values a bit. Also added an adjustment for energy gain for vore related actions. Red eyes are best at this, orange and purple are better at this than the other colors. Also made some adjustments based on the size you spawn as mostly for flavor.

-Fixed a couple runtimes that came up in testing.

---

-Prevents tesla, singulo, and explosions in general from affecting phased out shadekin.

---

-Removes the ability for shadekin to heal corpses with their healing abilities. - After consulting with medical players, this was determined to be a pain point, and considering that the ability was originally modeled after the strength of heal belly mode, that mode can not heal corpses, so, this probably shouldn't either.


crossexonar:
Phased shadekin on live seem able to press buttons that control doors or shutters, rascals pass kitchen shutters being  an example.

PR #789, opened 2025-02-18
Add the witch hat

TheGreatKitsune:
Adds the witch hat! By which I mean, I add it to the loadout since it was already present, just not a loadout option.


Lilly-Lira (approved):
Looks good to me!

PR #809, opened 2025-02-27
VChat 516 Compatibility

Arokha:
Adds 516 compatibility to VChat while maintaining 515 compatibility.
Shoutout to @Lilly-Lira for all the testing heeeelp.

The issues to be resolved revolved around WebView2's addition to 516, which supports modern JS.

I wrote VChat years ago when I barely knew JS. I think it was like my second JS project ever. I'd like to rewrite the whole thing in TypeScript honestly, with Vue 3 or some other framework, but I'll wait until 516 is stable and we can just dump all the 515 (i.e. Trident engine) support.

Key things changed:
- Created a small npm project for vchat to assist with minifying and eslint
- Convert some 'IE only' language (msMatchesSelector) into polyfill only used when necessary
- Fix 'double clicks' (a bug of Trident, apparently, not present in WebView2 that was being compensated for)
- Implement 'byondstorage' as an alternative on 516 clients to localstorage
- Byond's WebView2 has an (optional) implementation of devtools, so don't replace console.x if we don't need to

There was some other minor housekeeping but that's about it.


Lilly-Lira (approved):
Looks good!  All aspects seem to work fine in both 515 and 516

PR #810, opened 2025-02-27
20250227_1818_KH Remove 516 login warning.

lm40:
Once https://github.com/TS-Rogue-Star/Rogue-Star/pull/809 is merged merge this too, while 516 is still a beta I don't really think it will be necessary to constantly warn people of this


Lilly-Lira (approved):
With this, it seems we are fully compatible with 516.  The future is now :3

PR #836, opened 2025-03-11
Removes simple mobs as ID access when held.

crossexonar:
Disables you from picking up simple mobs that have access and using them as an ID card.

Use either this or https://github.com/TS-Rogue-Star/Rogue-Star/pull/837


Very-Soft (changes requested):
I would prefer this if it only applied while the mob doesn't have a ckey. It's not really a problem to have them have their access if they are being played.

Something more along the lines of

```
if(istype(I, /obj/item/weapon/holder))
  var/obj/item/weapon/holder/h = I
  if(!h.held_mob.ckey)
    return null
```

In this way, it would apply to micro holders too, and work nice and consistently~

PR #837, opened 2025-03-11
Removes carried mobs from being used as access pass, micros included.

crossexonar:
Disables you from picking up simple mobs that have access, or micros and using them as an ID card, ask the smalls to open door for you or take their ID.

Use either this or https://github.com/TS-Rogue-Star/Rogue-Star/pull/836

### DIFF
--- a/code/game/jobs/access.dm
+++ b/code/game/jobs/access.dm
@@ -218,6 +218,9 @@
 /mob/living/carbon/human/GetIdCard()
 	if(get_active_hand())
 		var/obj/item/I = get_active_hand()
+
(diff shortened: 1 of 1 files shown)


crossexonar:
> Looks like I will be throwing my friends at the doors instead...

No that was fixed years ago, please check the other version of this PR, I just put the code for both just in case.

Voidastra:
Bold of you to think I can read. But in all seriousness, I will give that a peek.

FartMaster69420:
But why?

crossexonar:
> But why?

Again, Im just posting Both versions so the code is on hand, the other one is more likely to be used.

PR #839, opened 2025-03-11
Fixes an issue with simple mob Put_in_hands()

### LINKED ISSUE #560: [GENERAL] Scugs delete IDs from PDAs
#### Brief description of the bug
While holding someone's PDA in hand while playing as a catslug, removing the ID from the PDA deletes it instead. For additional context, I was playing as Professor Nubbins, and stole a PDA from a scientist.

#### What you expected to happen
The ID should pop out of the PDA into the scug's open hand, or at least drop on the floor.

#### What actually happened
When specifically pressing the alt left click key to pop out the ID ( I have no idea if this bug occurs with the right click menu option) the ID simply deleted rather than being removed from the PDA.

#### Steps to reproduce
- (Step 1) Ghost and become Professor Nubbins
- (Step 2) Acquire crewmember PDA with an ID in it
- (Step 3) Alt click the PDA in one hand, with the other hand empty

#### Code Revision
- (Found using the "Show Server Revision" verb under the OOC tab.)
 

#### Anything else you may wish to add:
- (Location if it's a mapping issue, screenshots, sprites, etc.)
I did so because even while holding an ID'd PDA in hand and interacting with a vendor, catslugs cannot use the ID to purchase from the vendor which may be a separate but less important bug.

crossexonar:
Fixes simple mobs destroying items if they use functions similar to eject ID or Eject Pen with a full hand.

fixes #560

### DIFF
--- a/code/modules/mob/living/simple_mob/hands.dm
+++ b/code/modules/mob/living/simple_mob/hands.dm
@@ -11,9 +11,17 @@
 	return
 
 /mob/living/simple_mob/put_in_hands(var/obj/item/W) // No hands.
-	if(has_hands)
-		put_in_active_hand(W)
-		return 1
+	if(has_hands) //RS Edit Start | Fixes simple mobs destroying ID's from PDAs'
+		if(put_in_active_hand(W))
+			update_inv_l_hand()
+			update_inv_r_hand()
+			return 1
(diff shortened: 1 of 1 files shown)


Lilly-Lira (approved):
Did some testing and seems to work as intended!

PR #840, opened 2025-03-12
Sets pets from pet system to not be devourable, two line addition.

crossexonar:
" From the Council of PETA (People eating tasty animals) We notice that your pet system is violating the ethics code of eating non sentient Beings, specifically, we had previously made it so that you cannot consume station pets such as Ian, lisa, or keindrick after repeated issues, and have devised a firmware update for this system.
As a result, All pets, regardless of who they belong to, are no longer Consumable. " <- Meme funny lore reason text.

Literally just makes pets from the pet system not eatable, requested by Headmin Kenzie, please do not smite me.

### DIFF
--- a/code/modules/client/stored_mob.dm
+++ b/code/modules/client/stored_mob.dm
@@ -209,6 +209,7 @@
 	var/mob/living/simple_mob/M = new ourtype(get_turf(src))
 	M.mob_bank_load(user, load)
 	M.faction = user.faction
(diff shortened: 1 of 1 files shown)


CockatriceXL:
Does this apply if the pet has a CKey? Some pets are meant to be played, and some players who might play pets could be prey or switches.

crossexonar:
Another solution was found, awu.

PR #850, opened 2025-03-17
Fixes empty icon-states for several weapons

Michab02:
There were a few typos in the code. No idea how they got there. Let me know if anyone else finds any broken weapons like this.


crossexonar:
Wawa, you need to add your //RS edits / RS Fix
*pat woofer more

Lilly-Lira (approved):
Tested all three and work fine!

PR #852, opened 2025-03-18
Minor Pet adjustments

Very-Soft:
Makes slimes, nurse spiders, and broodmothers not devourable by default.

Makes slimes, nurse spiders, and broodmothers not pick-up-able by default. This will prevent them from being picked up by pet carriers.

Also makes nurse spiders and broodmothers not able to be saved to the pet system. This will not affect existing pets, only new pet registrations.

Also, allows people's pets to pass through the redgate, so they can more easily bring their pets on adventures.


Lilly-Lira (approved):
Looks good to me!

PR #853, opened 2025-03-18
adjust slime potions to not work on varedited mobs

crossexonar:
Prevents the slime potions that turn off mob AI from working on things that are var-edited.
Example: you have a boss dragon with adjusted Maximum HP, person walks up with a docility agent and ends the fight instantly.

### DIFF
--- a/code/modules/xenobio/items/slimepotions.dm
+++ b/code/modules/xenobio/items/slimepotions.dm
@@ -83,6 +83,9 @@
 	if(currently_using)
 		to_chat(user, span("warning", "This agent has already been used!")) // Possibly trying to cheese the dialogue box and use same potion on multiple targets.
 		return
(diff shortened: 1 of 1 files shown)


lm40:
It seems to me preferable to add a var that dictates whether this works or not, so you could varedit something and still have it be able to be docility agented if that's what you want for some reason but also that you can disable it if you don't want it

crossexonar:
> It seems to me preferable to add a var that dictates whether this works or not, so you could varedit something and still have it be able to be docility agented if that's what you want for some reason but also that you can disable it if you don't want it

Reason given for it to be this way is because apparently nobody would remember its even a thing if it was a variable,
I've got different opinions from you, lira, and soft, please come to a consensus on what I need to do with it.

PR #854, opened 2025-03-18
Port no touchy

crossexonar:
Ports and corrects a few issues with: https://github.com/VOREStation/VOREStation/pull/17289
Description: adds two traits "Reflexive biting" and "personal space" that allows you to toggle the ability to be:
Personal space: Prevents all touching, smelling, headpats, noseboops, etc, while active. 
Can be toggled on and off with "Toggle personal space"

Reflexive Biting: Prevents nose booping and headpats Specifically, causes 1 Brute damage and a funny bite message.
Can be toggled on and off with "Toggle Reflexive Biting"

### DIFF
--- a/code/__defines/species_traits.dm
+++ b/code/__defines/species_traits.dm
@@ -1,0 +1,7 @@
+//////flag defines specifically for species traits
+/// RS Port of Vorestation PULL #17289
+
+//touch reaction flags
+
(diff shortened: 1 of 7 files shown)


Skysteak:
This is not a big deal but it may be better for the 1 damage bite to do hallos instead. I feel that there may be some accidental ill-will if someone misclicks. There may also be potential escalation but then again if someone is going to do that over 1 damage...

PR #933, opened 2025-05-15
Smelly

Very-Soft:
Adds Olfactory Tracker as a positive trait available for all species to take.

Olfactory Tracker is activated by using the Smell verb. This will give the user a user interface element that will show the relative direction of the latest target that they smelled as long as they are on the same Z level. If the target is on a connected Z level, it will also let you know if the target is up or down from your location.

The tracking mode can be toggled by using the 'Track Target' verb found in the abilities tab. This will disable or enable the interface element.

When you begin tracking, a runechat message will appear for everyone, giving them the idea that you are sniffing around! It doesn't make a message in the chat log though, so you might be able to be a little sneaky about it if people aren't paying attention.

If one knows they are being tracked by smell, they can break all smell tracking by taking a shower or cleaning themselves with soap!

This also adds a couple of admin tools to allow staff to apply traits to people more easily in case that's a thing they want to do.


Lilly-Lira (approved):
Good trait :3

PR #936, opened 2025-05-16
real time

Very-Soft:
Makes the station time be the same as real time (eastern time)


Lilly-Lira:
Will the server have daylight savings time??

Very-Soft:
> Will the server have daylight savings time??

No

Lilly-Lira (approved):
Sad we must share our pain with them but approved

PR #943, opened 2025-05-21
Leopardmander and Big Dragon Adjustments

Very-Soft:
Leopardmander and Big Dragon have started with max nutrition, which before we multipled the max nutrition was fine, but now that the max nutrition is 50,000, they incidentally are really OP on power creation with the REGs.

These mobs had max nutrition however because of their use of healbelly, and healbelly consumes nutrition.

To make sure this is not compromised, slightly refactored how healbelly works, so that it won't consume nutrition from the mob if it has the natural_healer flag. This way these mobs can heal anyone as much as they want, and their nutrition will never be a factor.

Along with this, I decreased the leopardmander's human swallowing time to be in line with that of the bigdragon's healing process. It took so long that I thought it was broken, so, I made it so that it can swallow you before the weaken ends, which feels more correct.

### DIFF
--- a/code/modules/mob/living/simple_mob/subtypes/vore/bigdragon.dm
+++ b/code/modules/mob/living/simple_mob/subtypes/vore/bigdragon.dm
@@ -47,6 +47,7 @@ I think I covered everything.
 ///
 ///		Main type
 ///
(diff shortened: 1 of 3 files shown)


Lilly-Lira (approved):
Yes good

PR #958, opened 2025-05-26
Borgs can't look at machines

Very-Soft:
Makes it so that if borgs alt+click on a machine, then the look function won't fire. Borgs have existing alt+click functionality with a lot of machines, and how that's handled isn't the same across them. SO! Rather than adding custom handling for each of these, I think it's best to simply block the look function on machines for borgs!!!


Lilly-Lira (approved):
Wise

PR #969, opened 2025-06-05
Adds Mar Bar barsign by Alceris.

Alceris:
https://github.com/user-attachments/assets/61f3870d-77cd-4fa2-91b0-113104cc4fef


Very-Soft:
That's so cool!

Alceris:
I assume the failed check is irrelevant since the only thing this changes is a dmi file, which clearly does work since I did test it.

PR #983, opened 2025-06-25
damage overlay

Very-Soft:
Adds "Apply Damage Icons" addon option to bellies.

In RSUI, the icon displayed for the prey is the actual vis contents, which includes the damage overlay!

Now I have made it so that when considering adding the damage overlay while in a belly, it checks to see if that belly has this addon included before doing so. This means that, if they enter a hold belly with damage icons, they will probably keep them until something updates the damage overlay again! So, this mostly just affects digest mode and heal mode, as both of those will trigger it!

It should be noted that this affects the actual icon, not the vore panel icon. The vore panel icon doesn't update over time, so this won't change anything with that.

Also adds the "Vore Self Damage Overlay" preference. This starts enabled. When disabled, while in a belly damage overlays will not apply to your character's sprite. This is a global preference!


Lilly-Lira (approved):
Looks good!

PR #985, opened 2025-06-26
Re-Refixes My Kingside Borg Sprite

EtheoBoxxman:
The way it was read before was changed since my fix, which made my sprite use the wrong icon state. Changes it from jademanique-kingside-default > jademanique-kingside-rest so it works again.


Lilly-Lira (approved):
Looks fine!

PR #986, opened 2025-06-26
Port VOREStation PR17340 - Removes Baymiss

Durandaal:
Ports VOREStation [PR17340 ](https://github.com/VOREStation/VOREStation/pull/17340)and [PR17558](https://github.com/VOREStation/VOREStation/pull/17558).

This PR is a draft until I can test things more thoroughly and align with server intent, but the first brush is that it is immediately functional.

The full description can be found on the Virgo PR linked above, but highlights include that evasion is now calculated with the chance to hit; and should also extend to melee attacks. Attacks that make it through evasion also now have a 10% chance to hit a random limb, and point-blank shots will always hit players (within 2-3 tiles in early testing).

### DIFF
--- a/code/modules/mob/mob_helpers.dm
+++ b/code/modules/mob/mob_helpers.dm
@@ -167,22 +167,22 @@
 
 	//This is done here now, since previously it was just done in a dumb spot that made no sense.
 	//Even if you were Neo, anyone could land a blow on you.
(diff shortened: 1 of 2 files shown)


Lilly-Lira (approved):
Looks good to me!

PR #988, opened 2025-07-06
TGUI Duel Bundle Compatibility

Lilly-Lira:
Laying the groundwork to allow different TGUI bundles to coexist, with each interface able to pick the desired bundle.


TheCaramelion:
where contributions

PR #1015, opened 2025-07-27
Cleans empty newlines from circuit assembly examine text

Tenryumi:
Essentially a port of my Virgo PR, https://github.com/VOREStation/VOREStation/pull/18087 .

Before, whenever you examined a circuit assembly, it would print a new line for every single component in the assembly - EVEN if the component in question had nothing to show. This results in component-heavy builds spamming tons of empty newlines when they have nothing to show. Whoops!
This (hopefully) fixes that.

### DIFF
--- a/code/modules/integrated_electronics/core/assemblies.dm
+++ b/code/modules/integrated_electronics/core/assemblies.dm
@@ -213,7 +213,10 @@
 	. = ..()
 	if(Adjacent(user))
 		for(var/obj/item/integrated_circuit/IC in contents)
-			. += IC.external_examine(user)
+			// Make sure there's actually examine text to prevent empty lines being printed for EVERY component!
+			var/examine_text = IC.external_examine(user)
+			if (length(examine_text))
+				. += examine_text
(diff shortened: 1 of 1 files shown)


crossexonar:
Hello, given that this is a modification of non RS specific code, please add an //RS EDIT or //RS Add (whichever is applicapable) to the modified lines.
for large additions you may do something like 
// RS Add 
code giblets here
//RS Add End

Lilly-Lira:
If you could add the code comments, that would be great.  Otherwise I just learned circuit code to fix a bug a bit ago, so I'll try to test this out in the next couple of days so we can get it merged!

Tenryumi:
> Hello, given that this is a modification of non RS specific code, please add an //RS EDIT or //RS Add (whichever is applicapable) to the modified lines. for large additions you may do something like // RS Add code giblets here //RS Add End

Ah okay! Apologies, I didn't know.

Lilly-Lira (approved):
Works great!  Good addition to the code :3

PR #1047, opened 2025-08-25
Hunting tweak

Very-Soft:
Makes the various station pets (not from the pet system) not hunt!

Also makes it so semirandom mob spawners disable hunting by default.


Lilly-Lira (dismissed):
Looks good!

Lilly-Lira (approved):
Guaranteed to make Cocoa 90% less bloodthirsty.

PR #1071, opened 2025-09-14
New Painting System

Lilly-Lira:
Major update to the painting experience!

Lots of updates in this.  for painting, the brush now acts more like a brush, allowing you to make continuous strokes.  I also took some inspiration from the paint system Shadow did for Virgo (https://github.com/VOREStation/VOREStation/pull/16332), and added a line and a fill tool.  Beyond that, there is now an eyedropper for choosing a color off the canvas, and setting for setting brush size.  There is also a strength setting, allowing you to set the strength of the paint you are applying, so you can mix with the underlying color instead of completely replacing it.

There is also a new Entertainer job, Digital Artist, and a new tool, the Drawing Tablet.  The Digital Artist starts with the Drawing Tablet, and a couple spawn in the art room.  The drawing tablet is like a more advanced canvas.  In addition to the canvas features, it has addative and multiplicative shading options, an undo button, a clear canvas button, and three different layers to work with.  Shout out to Soft for providing input to provide features that are similar to an actual drawing tablet.

For the drawing tablet, you can print out the canvases, too!  You just need to use the tablet on a photocopier, and out will come a print!


lm40 (approved):
Didn't read but I'm sure it's fine

PR #1079, opened 2025-09-20
Trait Tweaks

DameonOwen:
- Allow Proteans and Prometheans to take autohiss traits.
- Refactor traits into species based trait maps:
  - Prevents traits that a species cannot take from appearing in the trait option list in the character creator (i.e. Xenochimera)
  - Reduces specific handling for `CUSTOM_SPECIES` during character sanitization
  - Fixes existing bug that allows users to take traits that their character's species shouldn't have.


Lilly-Lira (approved):
Works good!

PR #1140, opened 2025-10-21
Add missing APC cable in AI/Civilian Substation

DameonOwen:
The APC lasts all round without the cable, but it should probably be fixed.


DameonOwen:
Closing in favor of #1143

PR #1149, opened 2025-10-26
Mounted Flasher Fix

SadStarfish7:
To the dismay of any/all secmains, The mounted flashers don't actually activate! This PR fixes that.

Mounted flashers now look within a distance of two tiles for any standing, running, non-phased targets, and flashes them. They aren't hard to avoid (sunglasses, crawling, phasing, just walking), so they're pretty potent.

Some of this (lines 79-85) borrows from the code for handheld flashers, and the rest is me schizophrenically jumping through BYOND documentation after reading through the code for portable turrets.

Testing Evidence:

image
image
image
image
image

### DIFF
--- a/code/game/machinery/flasher.dm
+++ b/code/game/machinery/flasher.dm
@@ -56,11 +56,11 @@
 	last_flash = world.time
 	use_power(1500)
 
-	for(var/mob/living/O in oviewers(range,src))
+	for(var/mob/living/O in oviewers(range,src)) // RS EDIT
 		if(O.is_incorporeal())
 			continue
 
-		var/flash_time = strength // RS EDIT
+		var/flash_time = strength // RS EDIT START
 		if(istype(O, /mob/living/carbon/human)) // NIF Check!
 			var/mob/living/carbon/human/L = O
 			if(L.nif && L.nif.flag_check(NIF_V_FLASHPROT,NIF_FLAGS_VISION))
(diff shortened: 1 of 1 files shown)


code/game/machinery/flasher.dm
```diff
@@ -42,52 +40,59 @@
...
 		return
-
 	playsound(src, 'sound/weapons/flash.ogg', 100, 1)
 	flick("[base_state]_flash", src)
 	last_flash = world.time
 	use_power(1500)
 
-	for (var/mob/O in viewers(src, null))
-		if(get_dist(src, O) > range)
+	for(var/mob/living/O in oviewers(2,src))
```
Lilly-Lira: Lets change the 2 to instead be the range var so it stays dynamic with the variable.  Also add RS Edit code comments.

code/game/machinery/flasher.dm
```diff
@@ -97,25 +102,21 @@
...
 
-/obj/machinery/flasher/portable/HasProximity(turf/T, atom/movable/AM, oldloc)
+/obj/machinery/flasher/portable/process() // RS ADD
 	if(disable || !anchored || (last_flash && world.time < last_flash + 150))
 		return
-
-	if(iscarbon(AM))
-		var/mob/living/carbon/M = AM
-		if(M.m_intent != "walk")
+	for (var/mob/living/O in oviewers(2,src))
```
Lilly-Lira: Change to range and add code comments as above.

Lilly-Lira (approved):
Looks good!

PR #1163, opened 2025-11-23
Use Prettier (instead of Prettierx) for Style Enforcement

DameonOwen:
Uses the actively maintained and more supported `prettier` to perform style checks and make code more readable.
Also reduces the number of outdated dependencies being pulled in.


Lilly-Lira (approved):
Looks good!

PR #1166, opened 2025-11-24
Update TGUI minor dependencies

DameonOwen:
Performs minor version updates on a great many dependencies.
Additional testing help requested. I haven't found any oddities with these changes, and I don't expect anything to break with minor version bumps, but more eyes on it would be appreciated.


Lilly-Lira (approved):
Tested a bunch of windows and everything works as expected!

PR #1168, opened 2025-11-27
Resolve incompatability with BYOND 516.1648+

DameonOwen:
BYOND 516.1648 introduces a new `alist` associated list. As of this version of BYOND, normal lists cannot be an associated list with the key being a number.

This was not flagged in the compiler until 516.1670, instead resulting in runtimes. This PR updates the CI tool to use 516.1670 to highlight these issues. In addition, it solves existing issues found by the compiler in a backwards compatible way, so that the new `alist` is not used and the current server infrastructure can continue to be used.


Lilly-Lira (approved):
Looks good to me!

PR #1218, opened 2026-01-28
New portal just dropped

Very-Soft:
I also revamped the buttons again
I also refactored how redgates and event teleporters teleport you


Lilly-Lira (dismissed):
Looks good!

PR #1227, opened 2026-02-06
Join as mob restrictions

Very-Soft:
Prevents join as mob from being used in Central Command, Misc, and on the overmap.

Nothing most folk would notice, these are just places player mobs shouldn't be showing up unexpectedly~


Lilly-Lira (approved):
Also looks good!

PR #1313, opened 2026-04-11
Update CI

DameonOwen:
Update ahead of Node 20 EoL at the end of the month:

Update GitHub Actions agent to Ubuntu 22 (previous agent is no longer supported).
Update GitHub Actions to use Node 22.
Update _build_dependencies to point to Node 22.
Fix BYOND caching in CI.
Reduce calls to `install_byond` in CI.


Lilly-Lira (approved):
Looks good!

PR #1329, opened 2026-04-19
Blue space medigun, the final frontier

crossexonar:
Unified the name from "prototype bluespace medigun" to BSM-94 Bluespace medigun (ATM Machine lmao)

Gave a package/box to the BSM, click it in hands to unpack and drop contents on the floor.

Added the CMO version of this to the CMO locker finally, comes with a compact version upgrade (optional) and tier 2 parts (lets you see health of target like a basic analyzer, holds a tiny bit more, heals a tiny bit faster)

Made the BSM and its upgrade printable at science (does not come with preloaded power/chems), at the same tier/mats as the other medigun (Im unoriginal sue me)

tweaked a few bits of code here and there to make the thing work a bit better, and fixed the fact that it wouldn't heal burn at the same time as it did brute through a list shuffle.

Learned mapping thanks to autumwuff.

Most importantly: DOES NOT TOUCH THE ML3M IN ANY WAY. You may use whichever one you like (Please dont use both at once, I will cry, that is OP please do not)


code/modules/research/designs/medical_rs.dm
```diff
@@ -0,0 +1,26 @@
...
+/datum/design/item/medical/bluespacemedigun/AssembleDesignName()
+	..()
+	name = "Bluespace Medigun ([item_name])"
+
+/datum/design/item/medical/bluespacemedigun
+	name = "BSM-92"
+	id = "bluespace_medigun"
+	req_tech = list(TECH_MATERIAL = 6, TECH_MAGNET = 4, TECH_POWER = 3, TECH_BIO = 5)
+	materials = list(MAT_STEEL = 8000, MAT_PLASTIC = 8000, MAT_GLASS = 5000, MAT_SILVER = 1000, MAT_GOLD = 1000, MAT_URANIUM = 1000)
+	build_path = /obj/item/device/bork_medigun
```
Lilly-Lira: I think this is supposed to be continuous_medigun

code/game/Rogue Star/weapons/Medigun/linked_medigun.dm
```diff
@@ -236,6 +236,7 @@
 
 		for(var/name in BP_ALL)
 			var/obj/item/organ/external/O = H.organs_by_name[name]
+			shuffle(O.wounds)
```
Lilly-Lira: Instead I would update the for statement to for(var/datum/wound/W in shuffle(O.wounds))

Lilly-Lira (changes requested):
Two potential issues noticed.  More broadly, there is a third issue where the CMO version starts with a range of 3, even though it has the advanced scanner.  To get the range to go to five, the player needs to remove and then re-add the scanner.

code/modules/research/designs/medical_rs.dm
crossexonar: Fixed, will push in a few minutes (went to the protolathe and made sure it prints)

code/game/Rogue Star/weapons/Medigun/linked_medigun.dm
crossexonar: Testing if it still works the same without runtiming.

Lilly-Lira:
Don't have strong feelings on this one either way and just a thought, but should it require blue space tech levels?

crossexonar:
> Don't have strong feelings on this one either way and just a thought, but should it require blue space tech levels?

theyre available at round start anyway, same as the ml3m, if doctors really want more they can just deconstruct the one they have and print more, giving it additional tech would -probably- just make techweb easier/simple, considering how cracked researchers are at smashing out that list.

Lilly-Lira (approved):
Looks good!

PR #1331, opened 2026-04-22
Remove Who, rename Advanced Who to Who

Very-Soft:
## PR Purpose and Benefit

Comments out who, renames advanced who to who.

Both do the same thing, advanced who is just a better version of who, so I figured why not


Lilly-Lira (approved):
Looks good!

